PR #73, opened 2014-08-13
Add twig.so extension

buschjost:
Add twig extension

### DIFF
--- a/conf/php/conf.d/ext-twig.ini
+++ b/conf/php/conf.d/ext-twig.ini
@@ -1,0 +1,1 @@
+extension=twig.so
--- a/support/build/extensions/no-debug-non-zts-20121212/twig
+++ b/support/build/extensions/no-debug-non-zts-20121212/twig
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# Build Path: /app/.heroku/php/
+# Build Deps: php-5.5.11
+
+OUT_PREFIX=$1
+
+# fail hard
+set -o pipefail
+# fail harder
+set -eux
+
+DEFAULT_VERSION="1.16.0"
+dep_version=${VERSION:-$DEFAULT_VERSION}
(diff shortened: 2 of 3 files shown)


ceichinger:
+1

mikelohmann:
+1

djmattyg007:
@seb0zz because not everyone using this service is a Symfony2 developer?

buschjost:
@djmattyg007 the extension would not not be enabled for everyone. It would only be available to be enabled by users. So it has no impact to users who do not want to use it. 
Also there are many more users and frameworks using twig than just Symfony2.

seb0zz:
And even if only Symfony2 users would be using Twig? Symfony2 is one is the most used PHP frameworks and we are only talking about only giving the option to work with the Twig extension. It's not even that everyone HAS to use the extension. 

In fact it's the same with databases: Not everyone on the world uses MongoDB in their PHP Heroku projects but still they have the chance to enable the extension. Now Twig might not be such a critical part of an application as a database but still it's a very popular templating language and needs performance!

fabpot:
Having the Twig C extension by default would indeed be a great booster for any app using Twig. What do you think @dzuelke?

phansys:
:+1:

kbond:
+1

phansys:
@dzuelke, is there something missing in this PR to get merged? Please, let us know in order to do any required update.

seb0zz:
@dzuelke are you planning to give any feedback on community requests or do you want to keep ignoring them? -.-

dzuelke:
Will wait with merging at least until PHP 7 is supported in `ext-twig`. Can you, before or after that happens, please rebuild this PR using the new package infrastructure (see #104)?

buschjost:
I am sorry I will not be able to. We moved away from Heroku already.

PR #89, opened 2014-11-05
Allow running nginx without root, by changing default log output

nantunes:
nginx crashes when running with a unprivileged user because it can't write "/usr/nginx/logs/error.log". Specifing stderr stills gets a warning but is capable of continuing.


nantunes:
Closing as the PR is ignored.

PR #98, opened 2015-01-30
Add support for multi-version constraints

yukoff:
Consider following change to support composer multiple version constraints, eg. ~5.4.11|~5.5.0|~5.6.0


dzuelke:
This has now been fixed by adding escaping to the response on the API side.

PR #101, opened 2015-02-13
add pear mail and smtp for use with sendgrid/mandrill/etc addons

shwing:
There are a large number of buildpacks on GH that claim to do this, but they are not maintained. This is such a common feature that I believe it is worth considering it for the core buildpack.


dzuelke:
This is fundamentally against the modern principles of Composer. With this approach, you might get different versions of the PEAR packages on your local environment and on Heroku, which introduces bugs and headaches. That's one of the main reasons Composer exists in the first place.

Use https://packagist.org/packages/swiftmailer/swiftmailer or some other Composer package; they're more reliable, easier to install, and regularly maintained. PEAR is near dead.

PR #103, opened 2015-03-06
Anchor .php at the end of the filename

blongden:
Currently, this will rewrite *.phppppppp through fcgi. Unlikely, but surprisingly caused me an issue today :)

### DIFF
--- a/conf/nginx/heroku.conf.php
+++ b/conf/nginx/heroku.conf.php
@@ -63,7 +63,7 @@ http {
         }
         
         # default handling of .php
-        location ~ \.php {
+        location ~ \.php$ {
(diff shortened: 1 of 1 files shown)


dzuelke:
Does this prevent `index.php/pathinfo`?

blongden:
Good point. Perhaps change to `\.php([\/\?]?.*)$`

Icky regex, though.

blongden:
Oh and i'd need to test that...

dzuelke:
Can you test it please?

`.php($|/)` should work too.

dzuelke:
poke @blongden ? :)

dzuelke:
(no longer an issue)

Issue #104, opened 2015-03-17
Support installing custom PHP extensions without forking the buildpack and copying the whole bucket


stof:
This is something I already discussed with @dzuelke but I cannot find an issue tracking this, so here it goes.

The goal is to be able to install custom PHP extensions which are not provided by the buildpack itself.
There is 2 ways to provide this:

- easy (but slow) one: run pecl (or pickle) for these extensions
- faster (but more complex) one: allow apps to specify a place where the compiled extensions are available and then simply download them. This would also involve documenting the process to compile extensions for usage on heroku (and it needs to be able to support extensions needing a binding to an external lib btw)

martijngastkemper:
:+1:

dzuelke:
Yes, working on all of this at the moment. ETA is currently several weeks. It's gonna be a few "waves" of features, starting with changes to how configs are bundled, then support for versioned extensions, and finally the ability for multiple S3 repositories with binaries that are searched through sequentially, so you can have your own bucket of custom extensions (and there will be docs eventually of course).

slbmeh:
I just made a change to the compile in my fork to handle solution number 2.  It's in pr #111

slbmeh:
I just made a change to the compile in my fork to handle solution number 2.. it's in #110

stof:
@dzuelke any news ?

stof:
@dzuelke do you have any news about that ?

auro1:
status?

dzuelke:
All the necessary base work has been released for a few weeks. I am currently testing custom buckets and extension repos with someone. Seems to work really well. Should be able to roll it out next week.

NielsJanssen:
Exciting stuff, whats the status of this atm?

gboudreau:
@dzuelke Any news on this?

dzuelke:
It's all done and working. Just need to write docs and roll it out.

dzuelke:
(today!)

dzuelke:
https://github.com/heroku/heroku-buildpack-php/blob/develop/README.md is the new README, https://github.com/heroku/heroku-buildpack-php/blob/develop/support/build/README.md has initial build info (look through some formulae, such as "php" and "blackfire", for inspiration), and a crude Dockerfile lets you build without Heroku: https://github.com/heroku/heroku-buildpack-php/blob/develop/support/build/_docker/README.md

[...]

PR #110, opened 2015-05-15
add support for custom extensions downloaded from s3

slbmeh:
This allows custom defined extensions to be added in a forked buildpack and downloaded from an s3 bucket defined in the environment


dzuelke:
Closing in favor of #104.

PR #114, opened 2015-05-25
Support for post_compile hook

dz0ny:
This is same functionality as in https://github.com/heroku/heroku-buildpack-python


dzuelke:
First of all, there is no `sub-env` function defined in this buildpack, so that would fail miserably.

Second, this functionality already exists and is invoked a few lines above your change; documented here: https://devcenter.heroku.com/articles/php-support#custom-compile-step

PR #115, opened 2015-05-25
Updated README with buildpacks:set [skip ci]

jkutner:


dzuelke:
I'll actually remove that line altogether; makes more sense IMO. But going to re-use it in a different location, thanks :)

PR #118, opened 2015-06-25
$HTTP_RAW_POST_DATA automatic population removal

CTOJoe:
Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version of PHP. To avoid warnings, 'always_populate_raw_post_data' was set to '-1' in the php.ini configuration file.
http://php.net/manual/en/reserved.variables.httprawpostdata.php

### DIFF
--- a/conf/php/php.ini
+++ b/conf/php/php.ini
@@ -775,6 +775,7 @@ default_mimetype = "text/html"
 ; to disable this feature.
 ; http://php.net/always-populate-raw-post-data
 ;always_populate_raw_post_data = On
+always_populate_raw_post_data = -1
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Paths and Directories ;


phansys:
What is the use case for this PR? I ask because the [default PHP's behavior is to disable this feature](https://github.com/heroku/heroku-buildpack-php/blob/f3d631e6a6f43d8d5513bc228ab6f0ee55a980a9/conf/php/php.ini#L774-L775).

```ini
; Always populate the $HTTP_RAW_POST_DATA variable. PHP's default behavior is
; to disable this feature.
```

CTOJoe:
The default value is disabled but when running PHP version 5.6.10 with foreman every POST request appends the following in the log.
```
PHP 5.6.10 (cli) (built: Jun 23 2015 10:46:35) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
```

```
11:42:08 web.1    | started with pid 66396
11:42:08 worker.1 | started with pid 66397
11:42:08 web.1    | DOCUMENT_ROOT changed to 'public/'
11:42:08 web.1    | Using Nginx server-level configuration include 'nginx.conf'
11:42:08 web.1    | 4 processes at 128MB memory limit.
11:42:08 web.1    | Starting php-fpm...
11:42:10 web.1    | Starting nginx...
11:42:10 web.1    | Application ready for connections on port 5000.
11:42:13 web.1    | [23-Jun-2015 15:42:13 UTC] PHP Deprecated:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in Unknown on line 0
11:42:13 web.1    | 127.0.0.1 - - [23/Jun/2015:11:42:13 -0400] "POST /v1/accounts HTTP/1.1" 404 47 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.124 Safari/537.36
```

I assume the the warning is shown to bring awareness of the depreciation. The PR was only made to avoid the warnings polluting the syslog drains when supporting newer versions of PHP.

phansys:
I think it could be due you're using your local PHP settings. Try to run a [one-off dyno](https://devcenter.heroku.com/articles/one-off-dynos) and check this:
```
php -i | grep "always_populate_raw_post_data"
```

CTOJoe:
Foreman runs the Procfile:
```
web: vendor/bin/heroku-php-nginx -C nginx.conf public
worker: php artisan queue:listen
```
The Procfile starts the process type web which uses the buildpack's php.ini:
```
vendors/heroku/heroku-buildpack-php/conf/php/php.ini
```
I verified this on my local web.1, started with foreman, by performing a HTTP GET request to a PHP file containing phpinfo().

I understand that the default is 0. 

My local version of PHP's php.ini file also does not have 'always_populate_raw_post_data' declared and defaults to 0. 

When deployed to Heroku with the default PHP 5.6.10 runtime, the previously mentioned warning appears after each HTTP POST request.

This PR is to fix the warning when the buildpack is run using the PHP 5.6.10 runtime. It has been tested with both available runtimes of PHP 5.6.10 and PHP 5.5.26.

Ultimately I can override the setting explcitly by customizing my [php settings](https://devcenter.heroku.com/articles/php-support#customizing-settings-php), but I feel since it affects the default runtime it should be addressed.

phansys:
It is weird because I'm also using this buildpack with PHP ```5.6.10``` and Nginx, and I haven't this issue at all. Are you sure that the ```$HTTP_RAW_POST_DATA``` var isn't used in any fragment of your code?

CTOJoe:
We do not use $HTTP_RAW_POST_DATA in any of our source. The framework we use is [Laravel 5.1](http://laravel.com) which relies on composer for dependency management. I was unable to find any framework dependencies that use $HTTP_RAW_POST_DATA contained in the vendor directory created by composer.

phansys:
Just found this entry: https://bugs.php.net/bug.php?id=66763

CTOJoe:
You can experience the issue first hand with the following Heroku deployment:

[Deploy](https://heroku.com/deploy?template=https://github.com/laborsync/laravel/tree/heroku)


Once the app is live, tail its logs and you can use an app like [Postman](https://chrome.google.com/webstore/detail/postman/fhbjgbiflinjbdggehcddcbncdddomop?hl=en) to send it POST data. You should see the warning in the logs.

phansys:
After check with the app you provide as example, the same output is sent to client:
```
$ curl -X POST https://raw-post.herokuapp.com/
Check logs for:<br>"PHP Deprecated:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in Unknown on line "
```
I'll try to do some checks tomorrow.

CTOJoe:
That was by design for the example. The text in the HTML body was hardcoded in the app\Http\routes.php file as seen below:
```
Route::post('/', function(){
   return "Check logs for:<br>\"PHP Deprecated:  Automatically populating \$HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in Unknown on line \"";
});
```

Tail the logs of the app before posting to the root of the app domain. You will see the warning quoted in the post HTTP response body.

phansys:
Yes, I saw that too:
```
2015-06-27T00:12:55.209617+00:00 app[web.1]: [27-Jun-2015 00:12:55 UTC] PHP Deprecated:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in Unknown on line 0
```
But after reading in detail the link I shared before, I think this is the intended behavior for PHP 5.6 as the way to prepare for the complete removal of this feature in the next release. So, it should aim developers to explicitly set ```always_populate_raw_post_data``` to ```-1```.
So, if we force this setting in the buildback, we are imitating in PHP 5 the behavior which is intended for PHP 7, leaving off the possibility for users to get noticed about this change and make its own changes in consequence.

CTOJoe:
Good point. Maybe sweeping the intended warning message under the rug is not the proper solution. Perhaps instead the PHP docs should mention this warning and how it can be removed from logs using the supported [custom runtime settings](https://devcenter.heroku.com/articles/custom-php-settings#php-runtime-settings-ini-settings)?

phansys:
Yes, I agree that mentioning this could be helpful. Indeed, in a deprecation phase any extra help spreading the word and encouraging to follow the new rules isn't bad at all IMO.

dzuelke:
This warning is also triggered when you read from `php://input`; that's how it works inside PHP.

`always_populate_raw_post_data` is left at its default setting; one reason is that 5.5 can't accept a `-1`, the other is that I think it's best to use default settings for each release series wherever possible.

The idea is that, since we're all programmers, we'll read the deprecation notice, potentially do a bit of googling, and then follow the advice about "manually" setting `always_populate_raw_post_data` to `-1` if it bothers us :)

This is, also, of course not specific to Heroku. Any PHP 5.6 with a standard config will throw this.

dzuelke:
Closing this PR because `-1` is incompatible with PHP 5.5.

dzuelke:
This setting specifically, and how to change it, is also mentioned in the docs, by the way: https://devcenter.heroku.com/articles/custom-php-settings#php-fpm-configuration-include

PR #128, opened 2015-10-28
Add heroku.json support

EmanueleMinotto:
This PR is a proposal for an additional `heroku.json` file.

What's the problem? Composer currently doesn't support a `X or Y` requirement, so to work with packages like [Imagine](http://imagine.readthedocs.org/en/latest/) (afaik) there are three solutions:

 1. include `"ext-gd": "*"` or `"ext-imagick": "*"` in the composer.json/lock, this could be a problem if the library is included as a dependency
 2. edit project to add one of them in the composer.json, not so OSS-friendly, this forces the project to start on platforms with the GD extension already installed
 3. create a custom buildpack, not a long-term solution

This PR is to add to the project some Heroku only requirements.
Why are them in the `require` section? To allow a possible future usage of other root-level keys.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -263,6 +263,13 @@ if [[ "${PHP_VERSION}" ]]; then
         exts=($(cat "$COMPOSER" | python -c 'from __future__ import print_function; import sys, json; { print(key[4:].lower()) for key in json.load(sys.stdin)["require"] if key.startswith("ext-")}' 2> /dev/null || true)) # convert to array
         ext_source="$COMPOSER"
     fi
+
+    # adds Heroku only ext-* requirements
+    HEROKU_COMPOSER_SUPPORT="heroku.json"
+    if [[ -f "$HEROKU_COMPOSER_SUPPORT" ]]; then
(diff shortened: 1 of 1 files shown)


mmoreram:
+1

dzuelke:
I agree that this is a problem, but IMO it's something that should eventually be fixed in Composer by being able to express "either this or that extension" as a dependency.

I cannot merge this; the entire current process of runtime and extension selection is going away in two weeks as it won't work with PHP 7 any longer to pick a runtime first and extensions second; it all needs to be resolved through one dependencies graph based purely on `composer.lock`, and introducing new dependencies external to that will mean there are no longer guarantees that no conflicts occur.

Feel free to look at that change when it lands and see if you can fit your idea within that new system, but it'll likely be ranging somewhere between hard and impossible.

You can also have something like a `composer.heroku.json` with accompanying lock file, and set the `COMPOSER` env var using `heroku config:set` to that filename and we will use that file for builds instead if you just want different dependencies for Heroku altogether. Keep in mind the useful `--ignore-platform-reqs` switch for Composer's `install` and `update` commands as well.

PR #133, opened 2015-12-19
fix declared build deps on composer

stesie:
The composer recipe still references php-5.5.12 as build dep, however that one was removed recently.

This patch changes this to php-min


dzuelke:
Thanks, 6a8ec4b :)

PR #134, opened 2015-12-19
Add v8js.so extension

stesie:
This adds the v8js.so pecl extension, which allows embedding the V8 JavaScript engine into PHP


dzuelke:
Hey @stesie, I'll close this in favor of #104, which will allow you to maintain and host this yourself without actually running a forked buildpack. So far, there has not been a single request for `ext-v8js`; if demand ever becomes significant enough, we can revisit integration into the default extensions we provide.

PR #140, opened 2016-01-26
bcmath not enabled properly

bjyoungblood:


dzuelke:
Not sure what the point of this PR is... `--enable-bcmath=shared` means the extension gets built as a shared library ("`bcmath.so`") instead of statically into PHP. This is intentional.

If your `composer.json` properly declares your application's dependencies, including `ext-bcmath`, then that extension will be auto-enabled on deploy.

This is all documented at https://devcenter.heroku.com/articles/php-support#extensions

bjyoungblood:
My mistake. I could've _sworn_ that I had `ext-bcmath` in my composer.json, but apparently it was indeed missing.

dzuelke:
Always run `composer update` too to re-generate the lock file!

PR #150, opened 2016-02-17
Add a more helpful error message when attempting to install New Relic on HHVM

### LINKED ISSUE #149: New Relic extension won't work on HHVM
I've lost nearly a day on this, and I think it's probably my fault but it doesn't appear to be possible to install the built-in NR extension when using HHVM on the php buildpack.

```
remote: -----> Checking for additional extensions to install...
remote:        NOTICE: Using custom New Relic config 'newrelic.ini'
remote: /app/tmp/buildpacks/php/bin/util/newrelic.sh: line 53: cd: /tmp/build_251b7a28063dd325f9ef5b715f2cbd8a/.heroku/php/etc/php/conf.d: No such file or directory
```

After tearing my hair out, I remembered, **it's probably HHVM's fault.**

I'm assuming this file doesn't exist when using HHVM as the PHP runtime. As HHVM isn't officially supported by Heroku at this point (which is fine), might I suggest a more helpful error when attempting to install the NR extension on a HHVM app?

```
remote:  !     Push rejected, failed to install New Relic extension. Use PHP instead of HHVM.
```

Happy to submit a PR with changes to https://github.com/heroku/heroku-buildpack-php/blob/master/bin/util/newrelic.sh if someone can confirm this is the issue?

maxehmookau:
Fixes #149.

Currently, if you use HHVM as the runtime and also have the `NEW_RELIC_LICENSE_KEY` env var set, the build will fail with a very unhelpful error message. 

`remote: /app/tmp/buildpacks/php/bin/util/newrelic.sh: line 53: cd: /tmp/build_251b7a28063dd325f9ef5b715f2cbd8a/.heroku/php/etc/php/conf.d: No such file or directory`

This PR seeks to fail the build in this situation and provide a more helpful error message with potential solutions.

### DIFF
--- a/bin/util/newrelic.sh
+++ b/bin/util/newrelic.sh
@@ -4,6 +4,11 @@ install_newrelic_ext() {
     # special treatment for New Relic; we enable it if we detect a license key for it
     # otherwise users would have to have it in their require section, which is annoying in development environments
     NEW_RELIC_LICENSE_KEY=${NEW_RELIC_LICENSE_KEY:-}
+
+    if [[ "$engine" == "hhvm" && -n "$NEW_RELIC_LICENSE_KEY" ]]; then
+        error "New Relic does not support non-standard PHP runtimes ('${engine}'). You can either remove the NEW_RELIC_LICENSE_KEY variable or switch to a standard runtime."
+    fi
+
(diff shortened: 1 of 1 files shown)


maxehmookau:
My bash script is pretty crappy so feel free to suggest a way in which this could be better achieved.

dzuelke:
See #149

Issue #127, opened 2015-10-28
Any reason Gzip is commented out in nginx config?


redroot:
I'm looking to serve static assets with compression on a PHP site, and I noticed that the `gzip: on` command is commented out, any particular reason for this? I can't find any comments online or anything via git blame.

Any help would be appreciated, thanks in advance

holtkamp:
+1 on this, 

we've seen a significant increase in transfered content after migrating to Heroku from a server that did have GZip enabled yesterday 10-02-2016:
<img width="826" alt="screen shot 2016-02-11 at 09 40 26" src="https://cloud.githubusercontent.com/assets/776405/12972322/7f80c0ba-d0a4-11e5-8268-e2c9ab74e394.png">


<img width="845" alt="screen shot 2016-02-11 at 09 40 48" src="https://cloud.githubusercontent.com/assets/776405/12972326/879caa48-d0a4-11e5-828f-ed01b8ecbfbd.png">

PS, yeah we know, still got to reduce the overall amount of transferred content like scripts, but that is a different issue ;)

stof:
@dzuelke any comment on this ?

dzuelke:
Is it a problem to enable it in the custom config that 99% of Nginx users on Heroku need anyway? :)

dzuelke:
(it's not enabled in Apache either)

holtkamp:
In case it is not enabled by default, it might be useful to [extend the current documentation](https://devcenter.heroku.com/articles/http-routing#gzipped-responses) a bit? Maybe with examples?

On the otherhand, it might save Heroku considerable amount of data transfer by enabling it by default, so why not enable it? And it **is** 2016, right? :smile:

dzuelke:
Like those docs say, it's the responsibility of each application (and those docs are language agnostic, so no PHP specific stuff will go in there). So the normal Nginx or Apache configuration for compression applies.

Right now I see many customers only compressing JS/CSS/images, not page content. What would the default be? `gzip_vary on;`? What level? `gzip_disable` or not? Are all customers fine with `gzip_proxied any`? Will existing gzip setups break?

The defaults in Nginx are not enough. I can't just throw `gzip on;` into the default config; you need to set a whole bunch of directives...

holtkamp:
ok, guess that would be a justified reason as was requested by @redroot then...

dzuelke:
[...]

PR #159, opened 2016-04-21
Turn gzip on for nginx conf

RealLukeMartin:
Deploys kept failing after git pushed the tar to remote, would output this:

```
gzip: stdin: not in gzip format
tar: Child returned status 1
```

This commit fixed it.

### DIFF
--- a/conf/nginx/heroku.conf.php
+++ b/conf/nginx/heroku.conf.php
@@ -14,7 +14,7 @@ http {
     #keepalive_timeout  0;
     keepalive_timeout  65;
 
-    #gzip  on;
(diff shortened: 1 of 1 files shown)


RealLukeMartin:
Just saw this issue https://github.com/heroku/heroku-buildpack-php/issues/127

Closing this.

dzuelke:
The Nginx configuration has nothing to do with unzip errors during deploys; those happen because the connection to S3 sometimes drops unexpectedly.

RealLukeMartin:
Thanks for the explanation

PR #161, opened 2016-05-07
fix vendor-dir error messaging copypasta

jamesearl:
Error messages were incorrectly warning about a misdetected vendor-dir. Probably copypasta.


dzuelke:
Merged in 1f68be1, thanks!

PR #169, opened 2016-07-20
Improve the docker setup

stof:
Due to layer caching, apt-get update would not run again when changing the list of installed packages if they are in separate layers.

The second commit reads pip requirements from the requirements.txt file instead of duplicating them, making things easier to maintain (no need to change them in 2 places)

### DIFF
--- a/support/build/_docker/cedar-14.Dockerfile
+++ b/support/build/_docker/cedar-14.Dockerfile
@@ -3,8 +3,7 @@ FROM heroku/cedar:14
 WORKDIR /app
 ENV WORKSPACE_DIR=/app/support/build
 
-RUN apt-get update
-RUN apt-get install -y python-pip
+RUN apt-get update && apt-get install -y python-pip
 
 RUN pip install 'bob-builder>=0.0.10' 's3cmd>=1.6.0'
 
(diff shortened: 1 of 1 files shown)


support/build/_docker/cedar-14.Dockerfile
```diff
@@ -3,9 +3,10 @@ FROM heroku/cedar:14
...
-RUN apt-get install -y python-pip
+RUN apt-get update && apt-get install -y python-pip
 
-RUN pip install 'bob-builder>=0.0.10' 's3cmd>=1.6.0'
+COPY requirements.txt /app/requirements.txt
 
-ADD . /app
+RUN pip install -r /app/requirements.txt
+
+COPY . /app
```
dmathieu: Why change this from `ADD` to `COPY`?

dzuelke: @dmathieu: "For other items (files, directories) that do not require ADD’s tar auto-extraction capability, you should always use COPY.", see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy

dmathieu: 👍

dmathieu:
LGTM

stof:
@dzuelke @dmathieu ping

stof:
@dzuelke any news about this PR ?

PR #171, opened 2016-07-20
Improve error messages when running builds in a dyno and they cannot

stof:
this makes it easier to understand what went wrong

### DIFF
--- a/support/build/extensions/no-debug-non-zts-20121212/cassandra-bare
+++ b/support/build/extensions/no-debug-non-zts-20121212/cassandra-bare
@@ -24,7 +24,7 @@ dep_manifest=${dep_package}_php-$series.composer.json
 echo "-----> Building ${dep_package} (from PECL)..."
 
 # we need libgmp-dev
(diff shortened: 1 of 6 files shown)


stof:
when trying to access the details of the heroku/compliance check here, Changeling is asking me permissions to write to my personal info on github. This looks wrong to me.

dmathieu:
The compliance stuff is internal, though we're using it for public projects too.
Only herokai will have access to it anyway.

I'm not a big fan of how this is worded. We're ending up with quite a bit of semicolons in the same phrase.
How about:

```
Failed to install libgmp-dev. You need to run this in a docker build, not a dyno.
```

dzuelke:
Yeah I've already changed this in another way @dmathieu (need to run `apt-get update` first anyway, and gave that the error, then folded all `apt-get install` invocations into one).

dzuelke:
bdcdd594ebfe0e916a9a9cda97f118e4d420065a

stof:
@dmathieu you may want to add a description of changeling to explain that it is an internal heroku tool

PR #173, opened 2016-07-20
Use the JSON mime type when uploading JSON manifests to S3

### LINKED ISSUE #172: When uploading manifests the content type should be set properly
currently, JSON files are added in the bucket as ``text/plain``. It would be better to upload them with ``application/json``

stof:
The guessing algorithm of s3cmd detects them as text/plain.

Closes #172

### DIFF
--- a/support/build/_util/include/manifest.sh
+++ b/support/build/_util/include/manifest.sh
@@ -9,5 +9,5 @@ print_or_export_manifest_cmd() {
 }
 
 generate_manifest_cmd() {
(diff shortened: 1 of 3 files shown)


stof:
@dzuelke ping

stof:
@dzuelke any news about this PR ?

freeformz:
2 Comments:

1. You may also be able to use -M here which will guess the mime type based on file extension or "mime magic". Not sure if that will apply nicely in all cases and I prefer being explicit anyway.

2. The s3cmds look awfully alike, so you could possibly DRY them up a bit. Not sure it's actually worth it though.

With that said though it LGTM, :+1:

Issue #179, opened 2016-09-22
heroku-buildpack-php broken on macOS Sierra (10.12)


jussikinnula:
**Prequisities:**

- OSX updated to macOS Sierra (10.12)
- Xcode updated to latest
- Homebrew packages updated to latest
- Note that "bedrock-on-heroku" has it's own set of prequisities, like installing Homebrew packages for PHP, MySQL, etc

**Steps to reproduce:**

1. Clone https://github.com/frc/bedrock-on-heroku (WordPress "bedrock" installation, with deployment setup for Heroku)
2. Switch to the cloned directory
3. Install composer packages
4. Start Heroku Local

**Expected result:**

Heroku Local starts without any issues

**Actual result:**

Heroku Local fails to start, because heroku-buildpack-php (more specifically `vendor/bin/heroku-php-nginx` -script) is broken.

See further details (e.g. terminal output) from:
https://gist.github.com/jussikinnula/d2f93b64cc2feed52a614d3d5b6a47c3

JanneAalto:
This only applies when composer is installed from homebrew

JanneAalto:
fix https://github.com/heroku/heroku-buildpack-php/pull/180

PR #180, opened 2016-09-22
fix composer binary execution on macOS Sierra and composer installed …

JanneAalto:
…from homebrew

fixes https://github.com/heroku/heroku-buildpack-php/issues/179

### DIFF
--- a/bin/heroku-php-nginx
+++ b/bin/heroku-php-nginx
@@ -201,7 +201,7 @@ unset -f php-fpm # remove the alias we made earlier that would prevent newrelic 
 composer() {
     local composer_bin=$(which ./composer.phar composer | head -n1)
     # check if we the composer binary is executable by PHP
-    if file --brief --dereference $composer_bin | grep "bash" > /dev/null ; then # newer versions of file return "data" for .phar
+    if file --brief --dereference $composer_bin | grep -e "shell" -e "bash" > /dev/null ; then # newer versions of file return "data" for .phar
         # run it directly; it's probably a bash script or similar (homebrew-php does this)
         $composer_bin "$@"
     else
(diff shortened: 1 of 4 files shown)


jussikinnula:
@dzuelke works for me, thanks @JanneAalto! 👍

dzuelke:
Thanks for this. Turns out this doesn't just fix it for macOS!

Ubuntu 14.04:
> `a /usr/bin/env bash script, ASCII text executable, with very long lines`

Ubuntu 16.04:
> `Bourne-Again shell script, ASCII text executable, with very long lines`

And for the sake of documenting it:

Mac OS X 10.11:
> `a bash script text executable`

macOS 10.12:
> `POSIX shell script text executable`

freeformz:
Why not just:

```console
if [ -x $composer_bin ]; then
  $composer_bin "$@"
else
  php -n $composer_bin "$@"
fi
```

dzuelke:
Because `composer` could be (and commonly is) a PHAR archive (special `.php` package), and in that case, I want to run it through `php -n`, which prevents the loading of any config, so it will, on Heroku, also prevent New Relic from loading, sending bogus "CLI call" telemetry back, and outputting startup info into the logs.

bin/heroku-php-nginx
```diff
@@ -201,7 +201,7 @@ unset -f php-fpm # remove the alias we made earlier that would prevent newrelic
 composer() {
     local composer_bin=$(which ./composer.phar composer | head -n1)
     # check if we the composer binary is executable by PHP
-    if file --brief --dereference $composer_bin | grep "bash" > /dev/null ; then # newer versions of file return "data" for .phar
+    if file --brief --dereference $composer_bin | grep "shell" > /dev/null || file --brief --dereference $composer_bin | grep "bash" > /dev/null ; then # newer versions of file return "data" for .phar
```
freeformz: Try:
```console
if file --brief --dereference $composer_bin | grep -e "shell" -e "bash" > /dev/null; then # newer versions of file return "data" for .phar
```

freeformz:
@dzuelke explained why not to me. But also, see comment, you can use multiple `-e pattern`'s to grep.

dzuelke:
Please use `grep -e "bash" -e "shell"`, @JanneAalto, and make the same change in the other three boot scripts.

JanneAalto:
@dzuelke changed.

freeformz:
:+1: 

PS: Where are the tests?

PR #183, opened 2016-10-12
Use the content-hash from the lock file rather than the hash

stof:
Nightly builds of composer are not generating the hash anymore in their lock file: https://github.com/composer/composer/commit/8313e8687756059ef318a01386de2e3b789cebdd

### DIFF
--- a/bin/util/platform.php
+++ b/bin/util/platform.php
@@ -30,7 +30,7 @@ $require = [];
 if(file_exists($COMPOSER_LOCK)) {
 	$lock = json_decode(file_get_contents($COMPOSER_LOCK), true);
 	// basic lock file validity check
-	if(!$lock || !isset($lock["platform"], $lock["packages"], $lock["hash"])) exit(1);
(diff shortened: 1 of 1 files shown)


dzuelke:
Thanks. I need to make a bunch of other adjustments then to ensure the minimum version is 1.0 (custom PHARs), and throw out a few associated old workarounds such as the stdout and stderr redirects.

stof:
content-hash is generated in lock files since 1.0.0-alpha11. the change was done in September 2015: https://github.com/composer/composer/pull/4140

dzuelke:
Yeah I know but some customers are still building with `composer.phar` files present that are older than that, so I need to add a version check (to the boot scripts as well), and then I can remove a bunch of `tail -n 1` workarounds for older Composers that would spit out "outdated dev version..." warnings.

dzuelke:
Plus, there are a lot of projects out there that haven't been touched in a while and won't have that field, so their builds would then fail. I suppose the check should instead be for either of the two?

stof:
Well, we could indeed check for one or the other, and use the one being available. I will change this

stof:
@dzuelke updated

PR #184, opened 2016-10-14
Add -r flag to git rm --cached instructions

maxehmookau:
Currently if you accidentally commit and deploy a PHP project with the `vendor/` directory, you get a warning as you should.

However, it tells you to use `git rm --cached` but doesn't tell you to use the `-r` flag to recursively remove the directory from source control. This just updates the warning to something that works.


dzuelke:
Put the `-r` before the `--cached` please!

maxehmookau:
Done!

PR #200, opened 2016-12-22
Remove useless files

stof:
blackfire was the only extension having an unversioned file for each PHP version. All other extensions are referencing the ``no-debug-non-zts-20121212`` one directly from the unversioned file. So I did the same for blackfire for consistency.

### DIFF
--- a/support/build/extensions/no-debug-non-zts-20131226/blackfire
+++ b/support/build/extensions/no-debug-non-zts-20131226/blackfire
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-# Build Path: /app/.heroku/php/
-
-source $(dirname $0)/../no-debug-non-zts-20121212/blackfire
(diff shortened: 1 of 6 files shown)


dzuelke:
The purpose is to build whatever is the latest version without a Git change, in case someone wants to do that.

stof:
but then, why not for other extensions ?

stof:
Btw, the script has a big scary warning when doing it: https://github.com/heroku/heroku-buildpack-php/blob/960199a978308c75926fd9bb4775f7113bf1d777/support/build/extensions/no-debug-non-zts-20121212/blackfire#L46
So why allowing to do it ?

dzuelke:
Well... ;)

Explicit versioning is better, IMO (also because there is a cutover moment on S3 otherwise where manifest version and tarball version differ).

It's mostly legacy from when the Blackfire folks had written that script and it used their API to find the version. The logic for that is still in the "main" `blackfire` script.

dzuelke:
(in practice, most people, including Heroku itself, hopefully use a dev and a stable repo, and `sync.sh` between them, so the warning is not that relevant)

PR #215, opened 2017-02-21
add php artisan migrate to the end of the script

kosmonowt:


kosmonowt:
this was mistakenly opened this way. I wanted to sync my fork Oo

dzuelke:
You do not need to maintain a fork of the buildpack for this, @kosmonowt (it will occasionally get out of sync and then result in build failures).

There is a hook specifically for what you're trying to do: https://devcenter.heroku.com/articles/php-support#custom-compile-step

Also, keep in mind that this migration will not always run when you want it to, but only when an application is built. It will not run on simple config var changes (probably not an issue, but if you have e.g. two test DBs for a dev/staging app that you switch between using `heroku pg:promote`), or when promoting a built application in a pipeline e.g. from staging to production (https://devcenter.heroku.com/articles/pipelines). The Release Phase feature (https://devcenter.heroku.com/articles/release-phase) is designed specifically for database migrations.

kosmonowt:
Hi @dzuelke 
thanks for the tips.
I will use the release-phase and hope it's already stable enough for the migration script.

PR #217, opened 2017-03-01
chore(redirect): Added redirection for old url and old mobile users.

ErikSaunier:


dzuelke:
Wrong repo, @ErikSaunier, but you do not need to fork the buildpack for this anyway.

Just use the `-c` parameter (lowercase!) of `heroku-php-nginx` to supply your own "full" config.

Also, 2048 is not an allowed value for `map_hash_bucket_size`.

PR #223, opened 2017-04-10
Bugfix: allow heroku/heroku-buildpack-php to be in composer.json require-dev

gboudreau:
When compiling using `bin/test-compile` (like Heroku CI does), or whenever composer installs the dev dependencies (you can do that by setting the `HEROKU_PHP_INSTALL_DEV` environment variable to something that doesn't contain `--no-dev`), allow `heroku/heroku-buildpack-php` to be listed in the composer.json' `require-dev` section.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -248,14 +248,17 @@ fi
 # no need for the token to stay around in the env
 unset COMPOSER_GITHUB_OAUTH_TOKEN
 
+if [[ ${HEROKU_PHP_INSTALL_DEV-"--no-dev"} != *"--no-dev"* ]]; then
+    # dev dependencies will be installed; make sure those don't include heroku/heroku-buildpack-php
+    composer remove heroku/heroku-buildpack-php &> /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


gboudreau:
Of note: I'm not getting  the `Re: heroku/heroku-buildpack-php - This package may only be used as a dependency in 'require-dev'!` error anymore, but Heroku CI setup stills fails with error 82. Not sure if this is related.

dzuelke:
Thanks for this PR. I am currently working on a related change that installs the buildpack globally, outside the project tree, to avoid interference (it would still force-link the "boot scripts" as most folks have "`vendor/bin/…`" in their `Procfile`s).

Would that be a more forward-compatible fix, as it removes the underlying issue altogether?

gboudreau:
@dzuelke yes, maybe. Not sure what the underlying issue really is. :)

dzuelke:
I realized I never closed this issue after fixing it in 44ed2000

PR #225, opened 2017-04-16
Added support for Private Packagist

magnusnordlander:
This PR adds support for [Private Packagist](https://packagist.com), using two environment variables `PRIVATE_PACKAGIST_USERNAME` and `PRIVATE_PACKAGIST_TOKEN`. Setting both of these writes an auth.json file to $COMPOSER_HOME.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -74,6 +74,18 @@ mkdir -p $build_dir/.heroku/php
 export COMPOSER_HOME=$cache_dir/.composer
 mkdir -p $COMPOSER_HOME
 
+export_env_dir "$env_dir" '^PRIVATE_PACKAGIST_(USERNAME|TOKEN)$'
+if [[ -n ${PRIVATE_PACKAGIST_USERNAME:-} && -n ${PRIVATE_PACKAGIST_TOKEN:-} ]]; then
+    echo "{
+    \"http-basic\": {
+        \"repo.packagist.com\": {
+            \"username\": \"$PRIVATE_PACKAGIST_USERNAME\",
+            \"password\": \"$PRIVATE_PACKAGIST_TOKEN\"
+        }
+    }
+}" > $COMPOSER_HOME/auth.json
+fi
+
 # if the build dir is not "/app", we symlink in the .heroku/php subdir (and only that, to avoid problems with other buildpacks) so that PHP correctly finds its INI files etc
 [[ $build_dir == '/app' ]] || ln -s $build_dir/.heroku/php /app/.heroku/php
 


dzuelke:
I have a half-done branch, based on another half-done branch, locally that adds some general auth improvements. IMO there should be a more generic solution for auth; one option of course is to just provide a JSON string in `COMPOSER_AUTH`, but maybe the buildpack should also translate `COMPOSER_AUTH__HTTP_BASIC__COM_PACKAGIST_REPO=foo:bar` to JSON, or something like that?

dzuelke:
or, like Bundler, `COMPOSER_AUTH_HTTP_BASIC_REPO__PACKAGIST__COM`...

magnusnordlander:
I'm down with that.

Specifically, I agree that the buildpack should both support `COMPOSER_AUTH` and `COMPOSER_AUTH_HTTP_BASIC_<<domain>>` where ``<<domain>>`` is transformed to lower case and has ``__`` replaced with ``.``. I can implement this if you'd like. I also reckon that `COMPOSER_AUTH`, if defined would override all `COMPOSER_AUTH_*` variables?

Should I also implement support for `COMPOSER_AUTH_BITBUCKET_OAUTH_<<domain>>`, `COMPOSER_AUTH_GITHUB_OAUTH_<<domain>>`, `COMPOSER_AUTH_GITLAB_OAUTH_<<domain>>` and `COMPOSER_AUTH_GITLAB_TOKEN_<<domain>>` (the other allowed keys for auth.json)?

dzuelke:
`COMPOSER_AUTH` also being present is an interesting question. Which one takes precedence? ;) Or do we merge them? :<

The other alternative is to lobby Composer for support for such env vars. Bundler would do `BUNDLE_REPO__PACKAGIST__COM`, and we could, similarly, just go for `COMPOSER_REPO__PACKAGIST_COM` to cover the HTTP Basic auth case, although I don't really like that "simple" namespacing using a single prefix.

Either way, give me a few days so I can take a look at my "cleanup" branch and see if I can merge it. You basing your work on that would make more sense.

magnusnordlander:
Interestingly, Composer already supports a `COMPOSER_AUTH` environment variable.

Since @seldaek is probably interested in this anyway, he can consider himself pinged regarding Composer support for more specific auth environment variables à la the above. I'd be equally happy to work on 
a PR for composer/composer to accomplish this. Would you like me to?

dzuelke:
Ah, yes it does support that already, sorry, I thought you knew ;) (you can also use that in the meantime for Private Packagist auth)

Seldaek:
Just thinking out loud.. At the point where you have `COMPOSER_AUTH__HTTP_BASIC__COM_PACKAGIST_REPO=foo:bar`, isn't it easier to do `composer config --auth http-basic.repo.packagist.com foo bar`? I mean it's nice to support generic solutions but if they end up as verbose as the other generic solutions, what's the benefit?

PRIVATE_PACKAGIST_USERNAME at least is short and clear, but anyway yeah we could also support this in Composer I suppose, the question of what should take priority if COMPOSER_AUTH is also defined remains there too though.

dzuelke:
The purpose is to be able to set credentials using env vars for deploys on Heroku, @Seldaek. There is no "server" where you can run `composer config …`.

And with "`PRIVATE_PACKAGIST_USERNAME`", I'll then have to special case support for every new thing (Gemfury, for instance, or others).

Seldaek:
Right well anyway yeah I'm not really opposed to having this in Composer itself.

magnusnordlander:
@Seldaek: I'm working on a PR.

naderman:
@magnusnordlander hey, did you ever get around to this?

magnusnordlander:
@naderman: No, if I recall correctly, I got stuck on enumerating environment variables in a portable manner. The only way I could find to do this is by looping over `$_ENV`, but that only gets set if `variables_order` contains `E`. I still have the code so I can open a PR, but I'm not very happy with the implementation.

dzuelke:
Oh wow, @magnusnordlander, you're right. My PHP, out of the box, has no `E` in `variables_order`, so `$_ENV` is empty. Crazy.

I guess there is always `$_SERVER`... if both are unset, well, who does that? :p

magnusnordlander:
Yeah. This is the branch I was working in: https://github.com/composer/composer/compare/master...fervo:environment-auth

Thinking about it I did actually come up with an alternate approach though, but I haven't had the time to explore it yet. In my branch, because I don't know which domains and authentication types will be requested, I need to scan the entire environment. Obviously, at some point the code already knows which domain and authentication type it wants, though, so at that point, I should be able to narrow it down to checking a single, known, environment variable.

dzuelke:
Have you ever gotten around to this, @magnusnordlander?

dzuelke:
I'm closing this for now; we can't add a magic proprietary env var for every permutation of service and auth method (think about people's custom domain GitHub installs etc).

The docs at https://devcenter.heroku.com/articles/php-support#private-repositories have been updated to give more examples on how to simply use `COMPOSER_AUTH` for this purpose.

Maybe one day Composer will have built-in support for magic env vars (cheering you on, @magnusnordlander 🤗), but until then, a little bit of flattened JSON in an env var is not the end of the world IMO :)

PR #236, opened 2017-08-03
v122

dzuelke:


support/build/extensions/no-debug-non-zts-20131226/phalcon-3.2.1
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 # Build Path: /app/.heroku/php/
-# Build Deps: php-7.0.17
+# Build Deps: php-5.6.31
```
jkutner: This downgrade seems odd. just wanted to check that it's intended.

dzuelke: It is; thanks for checking. It was actually wrong before.

dzuelke: Ah sorry, never mind, it's just Git's rename detection throwing the paths around. These are actually different directories, so all is well.

PR #246, opened 2017-11-07
Set composer files explicitly

leshik:
The current buildpack version breaks WordPress Hybrid Composer plugin (it has composer.js file inside). Maybe it's better to explicitly exclude composer.json and composer.lock files.


dzuelke:
Thanks for your PR; I fixed it in eb35997 with proper terminating anchors.

PR #262, opened 2018-02-14
The PHP binary Composer uses should be equal to the binary the app uses. (Issue #261)

### LINKED ISSUE #261: Wrong version of libsodium installed (libsodium18 instead of libsodium23)
It appears the sodium library version is outdated.

This causes issues for applications that require newer functions (like built in Argon2 support)

Currently used version: `1.0.8` / `libsodium18`
Latest release: `1.0.16` / `libsodium23`

Example: symfony/symfony#26179

iSDP:
closes #261


iSDP:
Fixing this tomorrow, currently broken, could probably switch to the proper php binary after the installation of platform packages

stof:
The minimal pph version should still be used for the usage of composer used to resolve the installation of PHP itself.

all the code you are changing is related to the installation of platform packages, not to the installation of dependencies of your project (which already run with the proper PHP runtime AFAIK)

stof:
Switching to the proper binary after installing platform packages is *already* what the buildpack is doing currently.

PR #263, opened 2018-02-15
Updating libsodium to use the right version (libsodium23 over libsodium18)

### LINKED ISSUE #261: Wrong version of libsodium installed (libsodium18 instead of libsodium23)

iSDP:
fixes / closes #261

### DIFF
--- a/support/build/php
+++ b/support/build/php
@@ -46,7 +46,7 @@ fi
 # we need libgmp for GMP, libpam0g for IMAP, libicu for intl, libsasl2/krb/ldap for LDAP
 needed=( libgmp10 libpam0g $libicu libsasl2-2 libkrb5-3 libldap-2.4-2 )
 if [[ $STACK != "cedar-14" ]]; then
-    needed+=( libsodium18 )
+    needed+=( libsodium18 argon2 libargon2-0 libargon2-0-dev )
 fi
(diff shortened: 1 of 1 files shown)


support/build/php
```diff
@@ -109,6 +109,7 @@ if [[ $dep_version == 7.* ]]; then
     elif [[ $STACK != "cedar-14" ]]; then
         configureopts="\
```
stof: @dzuelke isn't this code loosing the `--enable-opcache-file` which was first added  for PHP 7+ above ?

iSDP: This was already there 🤔

stof: yeah, I know this is not related to your own change. Your own change seems good regarding existing code (if there is a bug, it was already there before)

iSDP:
Dangit, libsodium 1.0.8 is used, argon2 support was added in 1.0.9, current sodium version is 1.0.16

dzuelke:
The `libsodium23` package is not available in Ubuntu 16.04, which is what the Heroku-16 stack uses as its base image.

PR #265, opened 2018-02-16
feat: allow stack url overrides using BUILDPACK_VENDOR_URL env var

josegonzalez:
This is useful if you'd like to customize the PHP version without forking the entire buildpack.

Similar functionality was implemented in https://github.com/heroku/heroku-buildpack-ruby/pull/238

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -83,8 +83,16 @@ s3_url="https://lang-php.s3.amazonaws.com/dist-${STACK}-stable/"
 # prepend the default repo to the list configured by the user
 # list of repositories to use is in ascening order of precedence
 export_env_dir "$env_dir" '^HEROKU_PHP_PLATFORM_REPOSITORIES$'
+export_env_dir "$env_dir" '^BUILDPACK_VENDOR_URL$'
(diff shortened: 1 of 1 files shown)


josegonzalez:
Actually going to close. I didn't realize this wasn't what controlled where custom binaries live, and am actually not really sure _how_ I might override them since it looks like they are downloaded via composer somewhere...

dzuelke:
You don't have to fork the buildpack, @josegonzalez. In fact, only forking it gets you nowhere, because the vendored stuff of course lives in an S3 bucket.

You can add your own repositories (they're normal Composer repositories with custom package types), or override it entirely. It's all documented at https://github.com/heroku/heroku-buildpack-php/blob/master/support/build/README.md

Two examples of people using this: https://github.com/Incenteev/heroku-php-exts and https://tideways.io/profiler/article/35-installation-on-php#heroku

dzuelke:
(and to then use it on an app: https://github.com/heroku/heroku-buildpack-php/blob/master/README.md#custom-platform-repositories)

PR #267, opened 2018-03-04
Don't suppress FPM errors

rijnhard:
When using php in a frontend application it's relatively fine, but in an API context this is a much bigger issue.
Essentially what happens is nginx sees any output from php-fpm as valid, even though it's not, and will return 200 ok. 

[reference](https://ma.ttias.be/nginx-sets-http-200-ok-php-fpm-parse-errors/)

By default, it should fail, not provide a false sense of security.


dzuelke:
Those are already the defaults.

rijnhard:
then it would be because of something outside of the buildpack itself (I'm using them for another purpose) because I had to explicitly set them

dzuelke:
No idea if and where you're setting them; PHP's defaults are `display_errors = off` and `log_errors = on`, and we don't override them. I just double checked.

PR #269, opened 2018-03-21
log failure and warning counts

dzuelke:


bin/compile
```diff
@@ -16,8 +16,15 @@ mkdir -p "$cache_dir"
 env_dir=${3:-} # Anvil has none
 bp_dir=$(cd $(dirname $0); cd ..; pwd)
 
+export BPLOG_PREFIX="buildpack.php"
+export BUILDPACK_LOG_FILE=${BUILDPACK_LOG_FILE:-/dev/null}
+
+# stdlib
+source /dev/stdin <<< "$(curl -s --retry 3 https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh)"
```
schneems: I think we should be using `curl_retry_on_18` here.

bin/compile
```diff
@@ -139,12 +168,20 @@ echo "export PATH=\$HOME/.heroku/php/bin:\$PATH:\$HOME/$composer_bindir" > $buil
...
-/app/.heroku/php-min/bin/php $bp_dir/bin/util/platform.php "$bp_dir/support/installer/" $HEROKU_PHP_PLATFORM_REPOSITORIES 2>&1 >$build_dir/.heroku/php/composer.json | indent || error <<-EOF
-	Couldn't load '$COMPOSER_LOCK'; it must be a valid lock
-	file generated by Composer and be in a consistent state.
-	Check above for any parse errors and address them if necessary.
-	Run 'composer update', add/commit the change, then push again.
-EOF
+/app/.heroku/php-min/bin/php $bp_dir/bin/util/platform.php "$bp_dir/support/installer/" $HEROKU_PHP_PLATFORM_REPOSITORIES 2>&1 >$build_dir/.heroku/php/composer.json | indent || {
+	code=$?
+	if [[ $code -eq 3 ]]; then
+		mcount "failures.platform.composer_lock.runtime_only_in_dev"
```
schneems: You've got `failures.platform.composer_lock` and `warnings.composer_lock`. What's the significance of `platform`?

dzuelke: It's not a generic `composer.lock` problem, but only in the context of the "platform" (PHP+extensions+web server) installation.

PR #276, opened 2018-04-17
Support private packagist authentication

### LINKED ISSUE #275: Support private packagist
Similar to COMPOSER_GITHUB_OAUTH_TOKEN, it would be nice to have some way of logging in to private packagist instances during compile so composer can access private dependencies and/or use the private mirrors.

Something like:

```
composer config --global --auth http-basic.repo.packagist.com token $COMPOSER_PACKAGIST_TOKEN
```

neclimdul:
Add support for authenticating to a a private packagist instance.

Fixes #275 

https://packagist.com/

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -321,6 +321,15 @@ if [[ -f "$composer_vendordir/autoload.php" && -d "$composer_vendordir/composer"
 	EOF
 fi
 
+# handle custom packagist token
+export_env_dir "$env_dir" '^COMPOSER_PACKAGIST_TOKEN$'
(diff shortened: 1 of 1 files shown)


dzuelke:
I'm closing this for now; we can't add a magic proprietary env var for every permutation of service and auth method (think about people's custom domain GitHub installs etc).

The docs at https://devcenter.heroku.com/articles/php-support#private-repositories have been updated to give more examples on how to simply use `COMPOSER_AUTH` for this purpose.

Maybe one day Composer will have built-in support for magic env vars (see #225), but until then, a little bit of flattened JSON in an env var is not the end of the world IMO :)

PR #277, opened 2018-04-26
v135

hone:


dzuelke:
The redis extension is an intentional duplicate of v3 so people can run either if desired (v4 has breaking API changes)

PR #284, opened 2018-07-10
v138

dzuelke:


support/build/nginx
```diff
@@ -63,11 +63,11 @@ cp $(dirname $BASH_SOURCE)/_conf/nginx/nginx.conf ${OUT_PREFIX}/etc/nginx/nginx.
 mkdir -p ${OUT_PREFIX}/bin
 # this gets sourced after package install, so that the buildpack and following buildpacks can invoke
 cat > ${OUT_PREFIX}/bin/export.nginx.sh <<'EOF'
-export PATH="/app/.heroku/php/bin:/app/.heroku/php/sbin:$PATH"
+export PATH="/app/.heroku/php/sbin:$PATH"
 EOF
 # this gets sourced on dyno boot
 cat > ${OUT_PREFIX}/bin/profile.nginx.sh <<'EOF'
-export PATH="$HOME/.heroku/php/bin:$HOME/.heroku/php/sbin:$PATH"
```
jkutner: Are you sure `$HOME` doesn't need to be `\$HOME`. I'm suspect it's correct, since there was already a `$HOME` in there, but wanted to check anyways since I've made that mistake a bunch.

dzuelke: The `<<'EOF'` syntax prevents parameter expansion :)

PR #286, opened 2018-07-19
Add mime type for woff2 fonts

stof:


stof:
This cannot be added easily by using `-C` as it includes in the `server` context, and I'm not sure it can be done at this level.

stof:
@dzuelke ?

PR #154, opened 2016-03-11
Update buildpack image to align with marketing

mattgraham:
php

PR #289, opened 2018-07-25
run tests if 'composer test' is defined

dj95:
The buildpack for php should feature tests, like in the nodejs or ruby buildpack. As the nodejs features tests with `yarn test` or `npm test`the php-buildpack should feature tests run
with `composer test`.

This pull request contains a small test script, that executes the test command if it is definded.
If the command is not defined, a notice is printed.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+main() {
+    # parse the arguments into variables
+    declare build_dir="${1:-}"
+
+    # check if composer tests are available
+    if composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
+        # cd into the build directory and run the tests via composer
+        cd "$build_dir" && composer test
+    else # if no tests are defined...
+        # ...display a friendly reminder to define some
+        printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
+    fi
+}
+
+# fail hard
+set -eo pipefail
+# fail harder
+set -eu
+
+# execute the main function with command line arguments
+main "$@"
--- a/bin/test
+++ b/bin/test
@@ -13,7 +13,7 @@ declare build_dir="${1:-}"
 # check if composer tests are available
 if cd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
     # cd into the build directory and run the tests via composer
-    cd "$build_dir" && composer  run-script --timeout=0 test
+    cd "$build_dir" && composer run-script --timeout=0 test
 else # if no tests are defined...
     # ...display a friendly reminder to define some
     printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
--- a/bin/test
+++ b/bin/test
@@ -10,10 +10,10 @@ set -eu
 # name the command line argument for this script
 declare build_dir="${1:-}"
 
-# check if composer tests are available
-if cd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
-    # cd into the build directory and run the tests via composer
-    cd "$build_dir" && composer run-script --timeout=0 test
+# push the builddir to the directory stack and check if composer tests are available
+if pushd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
+    # run tests via composer in the build_dir
+    composer run-script --timeout=0 test
 else # if no tests are defined...
     # ...display a friendly reminder to define some
     printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
--- a/bin/test
+++ b/bin/test
@@ -1,23 +1,20 @@
 #!/bin/bash
 
-main() {
-    # name the argument to this function
-    declare build_dir="${1:-}"
-
-    # check if composer tests are available
-    if cd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
-        # cd into the build directory and run the tests via composer
-        cd "$build_dir" && composer test
-    else # if no tests are defined...
-        # ...display a friendly reminder to define some
-        printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
-    fi
-}
 
 # fail hard
 set -eo pipefail
 # fail harder
 set -eu
 
-# execute the main function with command line arguments
-main "$@"
+
+# name the argument to this function
+declare build_dir="${1:-}"
+
+# check if composer tests are available
+if cd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
+    # cd into the build directory and run the tests via composer
+    cd "$build_dir" && composer test
+else # if no tests are defined...
+    # ...display a friendly reminder to define some
+    printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
+fi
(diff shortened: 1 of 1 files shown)


Soolan:
Hi there,
Any update on this?
Tx

Soolan:
My pleasure @dj95 
It looks good to me, what I don't understand is why a simple PR like this takes forever to be merged.
We are still waiting for this issue to be resolved.

dzuelke:
@Soolan it's not up to you to approve PRs on a repository you do not own. I unfortunately cannot remove your approval as the GitHub UI won't let me.

Furthermore, it's not an issue that's blocking anybody in any way. You can put the `test` command into `app.json`, and everything will work just fine.

bin/test
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+main() {
+    # parse the arguments into variables
```
dzuelke: This comment is incorrect; it doesn't parse multiple arguments into variables.

bin/test
```diff
@@ -0,0 +1,23 @@
...
+    declare build_dir="${1:-}"
```
dzuelke: Why `declare`? You meant `local`?

dzuelke: `bin/test` doesn't receive arguments, so this whole bit is redundant anyway.

bin/test
```diff
@@ -0,0 +1,23 @@
...
+
+    # check if composer tests are available
+    if composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
+        # cd into the build directory and run the tests via composer
+        cd "$build_dir" && composer test
```
dzuelke: You could `cd ${1:-}` here directly, but see above

dzuelke: You probably want `composer run-script --timeout=0 test`

bin/test
```diff
@@ -0,0 +1,23 @@
...
+    else # if no tests are defined...
+        # ...display a friendly reminder to define some
+        printf "\033[0;31mNo test command found! Please define a test command in the composer.json in order to run test.\033[0m\n"
```
dzuelke: It needs to be discussed whether this should result in an error or not, given how users have to enable CI explicitly.

bin/test
```diff
@@ -0,0 +1,23 @@
...
+    fi
+}
+
+# fail hard
+set -eo pipefail
+# fail harder
+set -eu
+
+# execute the main function with command line arguments
+main "$@"
```
dzuelke: Please move the entire logic out of that function.

bin/test
josegonzalez: Looks like it does, based on the [nodejs buildpack](https://github.com/heroku/heroku-buildpack-nodejs/blob/master/bin/test).

bin/test
josegonzalez: I think the `cd` is correct, as noted above (and [here](https://github.com/heroku/heroku-buildpack-nodejs/blob/master/bin/test#L5)).

bin/test
josegonzalez: The `nodejs` buildpack doesn't check if there is a test command, so maybe this buildpack shouldn't either.

dzuelke: But that results in `npm` printing an error, yet still exiting with code 0. Other languages behave differently. And the underlying Heroku CI system will error the run (not fail it) if no tests are defined (and the buildpack doesn't have default test run behavior).

Soolan:
@dzuelke 
hmmmm let me see... it is not up to me yet I did it ;)
Sweetheart, go to settings and restrict outsider from approving. Believe me, it will look more professional than attacking ppl for approval =))

bin/test
josegonzalez: I was responding to this bin receiving arguments, not the declare vs local vs not assigning.

bin/test
josegonzalez: I think its worth just relying on the underlying toolchain. Its fairly easy for a user to set their test command to `true` if they decide that they do not want tests from their php buildpack to be invoked.

dzuelke:
The problem is not your approval, @Soolan (our internal compliance status check tool filters that), but the 

> It looks good to me, what I don't understand is why a simple PR like this takes forever to be merged.

comment.

And I'm not your sweetheart.

bin/test
dj95: I referred to the nodejs buildpack, like @josegonzalez wrote. The idea behind this declaration is that everyone who looks at the code exactly knows what the first argument of this function does. It's just for better readability.

`declare`is used in here, as `local`shifts the arguments. E.g.

```
main() {
    local arg1=$1 # would be $1 with declare
    local arg2=$1 # would be $2 with declare
}
```

bin/test
dj95: The check for the `composer test` command should be in done in order to not break existing CI pipelines. This warning may be irrelevant, but notifies the user to be able to test with `composer test`. But I can remove it, if you don't like the notification.

bin/test
dj95: I've done it in c7c79064f5da75e4a4c220520f9583132f4c0647

bin/test
josegonzalez: @dj95 for consistency's sake, lets just move to not having a function? I don't think thats the way any of the buildpacks are structured for bins (unfortunately) so introducing a new pattern isn't great :(

We can just add a comment at the top with a list of args etc. this thing is expecting.

dj95: I removed the function, as you can see in the new commits, and moved the declare statement to the top in order to keep the readability :)

dj95:
Do you have any updates on the recent commits? :)

bin/test
```diff
@@ -0,0 +1,20 @@
...
+set -eo pipefail
+# fail harder
+set -eu
+
+
+# name the command line argument for this script
+declare build_dir="${1:-}"
+
+# check if composer tests are available
+if cd "$build_dir" && composer run-script -l 2>/dev/null | grep -q '^\s*test\b'; then
```
josegonzalez: This could just be a separate:

```
pushd "$build_dir" > /dev/null
```

So you don't have to cd in the if statement.

dj95: Thank you for this hint! I think it's better to leave the pushd in the if-statement in order to check whether the script runs the test-command in an existing directory or not.

coudenysj:
Are there any plans on adding this (or an alternative) to the buildpack?

dj95:
I'm still waiting for more feedback in order to bring this pull request into a state that is acceptable for merging. If I get feedback, I'll implement it as soon as possible.

tortuetorche:
Hi @dj95,

Any news on this pull request?
This will be really useful with GitLab Auto DevOps feature, see https://gitlab.com/gitlab-org/gitlab-ce/issues/47651

Have a good day,
Tortue Torche

bin/test
scasei: if that error messages prevents this pr from beeing merged, than just remove it, please @dj95

scasei:
@dzuelke Last commit has closed the open discussion, what is left to get this PR merged?

dj95:
It's quite frustrating to see, how the maintainers care about feature requests and pull requests for this buildpack from the community. I mean they even didn't manage to get an updated logo merged for almost 2 years (https://github.com/heroku/heroku-buildpack-php/pull/154 - and this pull requests is from a heroku itself). Since there's no feedback or a sign that they will merge it, even if the code is well documented and only 20 lines long, I might close this issue in a few weeks as I don't see the point to offer this contribution when the maintainers do not want pull requests from the community.

Soolan:
@dj95 I can not agree more with you on this issue. This is just an example of irresponsible behavior from the repo maintainers.

dzuelke:
Please see #350 for a merged implementation.

PR #291, opened 2018-08-07
v141

dzuelke:


bin/detect
```diff
@@ -1,7 +1,29 @@
...
 
+# convenience functions
+source "$(cd $(dirname $0); cd ..; pwd)"/bin/util/common.sh
+
 if [[ -f "$1/composer.json" || -f "$1/index.php" ]]; then
 	echo "PHP" && exit 0
 else
-	exit 1
+	error <<-EOF
+		Application not supported by this buildpack!
```
jmorrell: This message is great, and I'm totally going to steal it!

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # heroku-buildpack-php CHANGELOG
 
+## v141 (2018-08-07)
```
jmorrell: This should really be several individual PRs

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+### ADD
+
+- ext-redis/4.1.1 [David Zuelke]
+- ext-mongodb/1.5.2 [David Zuelke]
+
+### CHG
+
+- Verbose error messasge on `bin/detect` failure [David Zuelke]
+- Emit brief warnings for common regexed build failure cases [David Zuelke]
+- Run most internal 'composer' invocations using '--no-plugins' [David Zuelke]
```
jmorrell: Why? Can you link this to an issue?

bin/compile
```diff
@@ -314,6 +297,35 @@ composer validate --no-check-publish --no-check-all --quiet "$COMPOSER" 2>/dev/n
 	EOF
 }
 
+# if prefer-stable is false and minimum-stability is not stable, warn about potential unstable platform installs
+[[ -f "$COMPOSER_LOCK" ]] && cat "$COMPOSER_LOCK" | python -c 'import sys, json; l = json.load(sys.stdin); sys.exit(l.get("minimum-stability", "stable") != "stable" and l.get("prefer-stable", False) == False);' 2> /dev/null || {
+	mcount "warnings.composer_lock.minimum_stability"
+	warning <<-EOF
+		Non-stable 'minimum-stability'!
```
jmorrell: This warning reads better without this first line I think. The next paragraph is a better place to start.

bin/compile
```diff
@@ -314,6 +297,35 @@ composer validate --no-check-publish --no-check-all --quiet "$COMPOSER" 2>/dev/n
...
+	
+		Your '$COMPOSER' contains a 'minimum-stability' lower than
```
jmorrell: Would printing what they do have set for `minimum-stability` be useful here? Maybe show that snippet of the config?

bin/compile
dzuelke: The "heading" is intentional; the first line is re-printed in a warnings summary in case an error occurs.

jmorrell: Then I'd suggest changing `Non-stable 'minimum-stability'!` to something a little more friendly? 🤔 

`Configuration could lead to using unstable versions of PHP` maybe?

CHANGELOG.md
dzuelke: There is an issue link in the commit message and above the first line that uses `--no-plugins` :)

bin/compile
dzuelke: It's supposed to be unfriendly, as it could result in people getting really unexpected stuff. This message used to show up for anyone with `minimum-stability` other than `stable`, regardless of `prefer-stable`, which is no longer the case, so the users who now still see it really need to act on it, otherwise they'll get a PHP 7.3 RC once it's out.

bin/compile
dzuelke: Done!

PR #305, opened 2018-10-31
Document the current state-of-the-art for PHP add-on provider support

djcp:
@dzuelke et al.

Please feel free to correct my language, tone, and documentation of the process here. I want to have an authoritative place to send add-on providers looking to support PHP with a binary extension in the most elegant way possible.


dzuelke:
Done in #331!

PR #306, opened 2018-11-05
v146: Install buildpack to .heroku/php/

dzuelke:
This avoids cluttering the user-owned vendor/ folder with our buildpack system installation, and we could in the future potentially allow users to specify their own buildpack fork/version for runtime stuff as a dependency in composer.json (right now, bin/compile still throws an error if the buildpack is detected).

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -347,7 +347,7 @@ else
 fi
 export HEROKU_PHP_DEFAULT_RUNTIME_VERSION
 # extract requirements from composer.lock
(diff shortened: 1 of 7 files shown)


bin/compile
```diff
@@ -704,18 +704,13 @@ fi
 
 status "Preparing runtime environment..."
 
-# install this buildpack like a composer package
+# this buildpack got installed as a composer package in the platform install step
```
jmorrell: `got installed as` -> `was installed as`

dzuelke: fixed

PR #310, opened 2018-12-13
v147

dzuelke:


support/build/php
```diff
@@ -200,9 +203,14 @@ MANIFEST_EXTRA="${MANIFEST_EXTRA:-"$extra"}"
 # remove temporary ini file that enables all extensions, and the composer download
 rm ${OUT_PREFIX}/etc/php/php.ini composer.phar
 
-cp $(dirname $BASH_SOURCE)/_conf/php/php.ini ${OUT_PREFIX}/etc/php/
-cp $(dirname $BASH_SOURCE)/_conf/php/conf.d/*.ini ${OUT_PREFIX}/etc/php/conf.d/
-cp $(dirname $BASH_SOURCE)/../../conf/php/php-fpm.conf ${OUT_PREFIX}/etc/php/
+IFS='.' read -r -a version <<< "$dep_version" # read the parts of $dep_version into an array, e.g. (7 2 33)
```
jmorrell: This is the kind of thing I wish there were tests for :( It can be as simple as naming a function and writing unit tests. Here's the example from the Node buildpack: 

https://github.com/heroku/heroku-buildpack-nodejs/blob/master/test/unit

bin/compile
```diff
@@ -218,7 +218,7 @@ fi
 mkdir -p $build_dir/.heroku/php-min
 ln -s $build_dir/.heroku/php-min /app/.heroku/php-min
 
-curl_retry_on_18 --fail --silent --location -o $build_dir/.heroku/php-min.tar.gz "${s3_url}php-min-7.0.32.tar.gz" || {
+curl_retry_on_18 --fail --silent --location -o $build_dir/.heroku/php-min.tar.gz "${s3_url}php-min-7.0.33.tar.gz" || {
```
jmorrell: Please double check to make sure these urls exist before deploying

support/build/php
```diff
@@ -102,7 +104,8 @@ fi
 configureopts=()
 if [[ $dep_version == 7.* ]]; then
 	configureopts+=("--enable-opcache-file")
-
+	configureopts+=("--without-libzip") # FIXME: change to --with-libzip once cedar-14 is EOL, as the other stacks have an up to date enough version
```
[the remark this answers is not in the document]
dzuelke: It means it uses the libzip bundled with PHP (`cedar-14` doesn't have it). Once cedar-14 is EOL, the builds will be switched to use the stack libzip.

support/build/php
dzuelke: I agree in principle, but I should mention that this is a build formula. It compiles PHP. Unit testing that is a little... well... :) The more appropriate test would likely be an integration test that verifies the resulting build uses the right configs.

support/build/php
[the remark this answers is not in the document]
jmorrell: ohh, I see. It works the opposite of how I expected the flags to work "build in env without libzip" vs "dont include libzip"

bin/compile
dzuelke: Of course. That's part of the rollout process :)

ajoy39:
something is slightly off here, I see you added the latest newrelic agent and that should in support php 7.3 but it's not working for us 

https://github.com/heroku/heroku-buildpack-php/issues/312

PR #329, opened 2019-03-13
Nginx/1.14

dzuelke:


stof:
@dzuelke why using 1.14 rather than 1.15 ? According to https://www.nginx.com/blog/nginx-1-6-1-7-released/ the official recommendation of the Nginx team is to use the mainline releases, unless you are concerned with compatibility with third-party modules. And the buildpack does not use any third-party module AFAICT.

dzuelke:
Mostly the tradeoff stability/effort versus benefits, @stof; the mainline occasionally has more severe bugs (e.g. 1.15.4/1.15.5, although that particular one would not have hit users), and with monthly-ish releases, it increases the workload (releasing these things does come with effort around testing, review and so forth) for little gain: most of the stuff in 1.15 is around SSL, TLS1.3, proxies, or more exotic modules, all of which is of little to no use on Heroku.

The generated version selector for the platform `composer.json` has been changed from `~1.8.0` to `^1.8.0` with this PR though, so you can build your own 1.15 package if you'd like, and it'll get installed, no buildpack fork needed :)

stof:
Well, I don't have a specific need for running 1.15 rather than 1.14. that was mostly curiosity.
I'm curious about when Heroku will support TLS 1.3 as well, but upgrading nginx in my app won't help me on that front :smile:

Issue #348, opened 2019-07-24
Broken on Fedora 29 PHP/Composer - calling composer with not php.ini


martin-langhoff:
On Fedora 29, using Fedora's PHP and Composer, with heroku buildpack v158 (according to CHANGELOG.md)...

```
$ /usr/local/bin/heroku local 
[OKAY] Loaded ENV .env File as KEY=VALUE Format
11:33:33 AM web.1 |  AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using fe80::cbd0:5dda:5580:ed9f. Set the 'ServerName' directive globally to suppress this message
11:33:33 AM web.1 |  Unable to determine Composer vendor-dir setting; is 'composer' executable on path or 'composer.phar' in current working directory?
[DONE] Killing all processes with signal  SIGINT
11:33:33 AM web.1 Exited with exit code null
```

Exact versions of Fedora components
```
$ rpm -qa php* composer*
...
php-fedora-autoloader-1.0.0-4.fc29.noarch
...
php-cli-7.2.20-1.fc29.x86_64
php-common-7.2.20-1.fc29.x86_64
...
php-fpm-7.2.20-1.fc29.x86_64
php-symfony-finder-2.8.51-1.fc29.noarch

$ rpm -qf /usr/bin/composer
composer-1.8.5-1.fc29.noarch

```

What I suspect happens is that Fedora is using its php.ini to manage the autoloading. So calling `/usr/bin/composer` works, but calling `php -n /usr/bin/composer` breaks. See below:

```
$ /usr/bin/composer config vendor-dir 
vendor
$ php /usr/bin/composer config vendor-dir 
vendor
`$ php -n /usr/bin/composer config vendor-dir 

Fatal error: Uncaught Error: Call to undefined function Composer\Json\json_decode() in /usr/share/php/Composer/Json/JsonFile.php:153
Stack trace:
#0 /usr/share/php/Composer/Factory.php(290): Composer\Json\JsonFile->validateSchema(1)
#1 /usr/share/php/Composer/Factory.php(576): Composer\Factory->createComposer(Object(Composer\IO\ConsoleIO), './composer.json', false)
#2 /usr/share/php/Composer/Console/Application.php(345): Composer\Factory::create(Object(Composer\IO\ConsoleIO), NULL, false)
#3 /usr/share/php/Composer/Console/Application.php(458): Composer\Console\Application->getComposer(false, false)
[...]

PR #349, opened 2019-07-24
User platform composer correctly when file utility says "executable"

martin-langhoff:
Addresses #348 --

On Fedora 29
```
$ file $(which composer) 
/usr/bin/composer: a /usr/bin/env php script, ASCII text executable

```

### DIFF
--- a/bin/heroku-hhvm-apache2
+++ b/bin/heroku-hhvm-apache2
@@ -235,7 +235,7 @@ httpd_version="$(httpd -v | hhvm --php -r 'echo preg_replace("#^Server version: 
 composer() {
 	local composer_bin=$(which ./composer.phar composer | head -n1)
 	# check if we the composer binary is executable by PHP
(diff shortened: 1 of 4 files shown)


dzuelke:
Thanks for this PR.

This fix is actually in the wrong place, as it now causes a PHAR file to be handled the same way as a Homebrew style wrapper shell script. This would break people with e.g. a local `composer.phar` that's not marked with an executable flag.

Thanks for this report.

The problem is actually that Fedora (and others) compile all extensions as shared, including e.g. JSON and PCRE, and then load them using individual INIs placed into the `--with-config-file-scan-dir` location.

Using `php -n` prevents all INIs from being loaded, so the extensions don't get loaded because that directory isn't scanned either.

Can you please try branch `php-dash-n-refactor`, see PR #351, and report whether that works for you?

dzuelke:
Alrighty, I have been able to confirm that #351 fixes this! Merging to master soon; a tagged release will come early next week.

PR #350, opened 2019-08-01
Heroku CI improvements: zend.assertions=1 and bin/test

dzuelke:
This PR:

1. adds a bunch of CI tests
1. enables [`zend.assertions`](https://www.php.net/manual/en/ini.core.php#ini.zend.assertions) for Heroku CI (assertions are otherwise disabled on Heroku, which is the recommended production PHP default)
1. adds a `bin/test` that supports both the quasi-standard `composer test` script, as well as all major test frameworks


jkutner:
I didn't look closely at the fixtures, but +1 otherwise

PR #355, opened 2019-08-08
test performance improvements

dzuelke:
try and fail faster, be verbose where needed and not where not, and properly weigh the parallelized test runs, which gets the runtime down to ~10 minutes across all stacks


test/var/log/README.md
```diff
@@ -0,0 +1,7 @@
+The `parallel_runtime_rspec.*.log` files in this directory are here on purpose.
```
jmorrell: Oof, splitting them up manually like this makes sense as a work-around, but this is something that should be handled by the testing framework with a queue / workers 😢

I've been considering something similar to help get Node CI times down

.travis.yml
```diff
@@ -8,9 +8,11 @@ install:
 - bundle install
 before_script:
 - bundle exec hatchet ci:setup
-script: bundle exec parallel_rspec -n6 test/spec/
+- heroku update
+script: bundle exec parallel_rspec -n 7 --runtime-log "test/var/log/parallel_runtime_rspec.${STACK}.log" --verbose-process-command --combine-stderr --prefix-output-with-test-env-number test/spec/
```
jmorrell: I don't understand this, but will assume you know what you're doing

test/var/log/README.md
jmorrell: Just to be clear, I'm expressing frustration at this problem myself. This is not saying you shouldn't do this

.travis.yml
dzuelke: There's four separate commits in this PR that all touch this single line, each with an explanation, if you're curious :)

test/var/log/README.md
dzuelke: Yea. RSpec :( You don't even need external queues or workers really for Hatchet, as it's 99% I/O wait. If `parallel_tests` didn't pre-assign spec files to the processes it spawned, but instead let them work one by one, none of this would even be necessary.

PR #359, opened 2019-08-16
Added support for GitLab token

Alexwijn:
I'm using this piece of code for like 8 months without any problems. I think I find it usefull to have it merged in the base repository.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -600,7 +600,34 @@ if [[ -f "$composer_vendordir/autoload.php" && -d "$composer_vendordir/composer"
 	EOF
 fi
 
-# handle custom oauth keys
+# handle gitlab token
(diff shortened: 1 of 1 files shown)


StefanoGroenland:
Yup, i can verify that we run fine with our fork. Would be great if we could switch back to the original repo with this addition!

dzuelke:
We won't be merging additional custom auth env var support, as it simply clutters the code too much.

You can instead use the `COMPOSER_AUTH` environment variable; please check the detailed documentation at https://devcenter.heroku.com/articles/php-support#private-repositories

The special variable for GitHub pre-dates `COMPOSER_AUTH`, so it will remain for backwards compatibility, but it's no longer documented.

PR #368, opened 2019-10-01
[changelog skip] Fix Escaping in Changelog Script

schneems:
The previous PR had a bug where the REGEX for grep was not properly escaped. This PR fixes that issue.


edmorley (changes requested):
Could you update this to use the newer version of the check? eg like:
https://github.com/heroku/heroku-buildpack-ruby/blob/main/.github/workflows/check_changelog.yml

edmorley:
Closing since this is stale.

PR #381, opened 2020-01-23
v167

dzuelke:
This PR contains three changes related to dyno startup and termination:

1. The leader processes (`heroku-php-apache2` etc) will now always attempt to gracefully shut down their child processes to allow requests to finish.  
PHP-FPM and Nginx will be sent a SIGQUIT, and Apache will be sent SIGWINCH; all of these signal "finish current requests and then shut down".  
This will prevent currently running requests from getting interrupted during a restart.
This is not yet widely useful, except with an experimental feature flag on the platform that only SIGTERMs the leader process instead of the entire process group.
2. The boot scripts will dynamically poll for `php-fpm`'s PID file as a way of determining whether the server is ready, versus waiting for a blanket two seconds to give it time to initialize.  
A bit of a wait is necessary, as in particular Nginx will start up more quickly than FPM, and if a request arrives immediately, the FPM backend wouldn't be ready and Nginx would mark the backend as unhealthy permanently.  
This change cuts the delay from two seconds down to fractions of a second.
3. The New Relic daemon `.profile.d` startup script will dynamically poll for `newrelic-daemon` to have finished startup, versus waiting for a blanket two seconds.  
Same effect as in 2): fractions of a second delay instead of two seconds.

Also cleans up New Relic logging (no longer need a log file we then have to tail), and fixes a few test related matters.

To review this PR, go commit by commit, and for the first commit, focus on just one file (e.g. `bin/heroku-php-apache2`), as the changes are all identical in the four scripts (yes, sharing code between these is on the todo, but non-trivial for various reasons).


hone:
@Malax Slack is great for discussions, but it has a major downside since
it's really a black hole of information. I treat slack like speaking to
somewhere in person for historical archiving. If there's anything critical,
it'd be great to get a summary back on the PR.

Malax:
@hone fair point! I just went through the Slack history and there is nothing critical I think. I asked about the intent behind this change and clarification what certain commands do when my google-fu failed me.

PR #385, opened 2020-02-06
PHP 7.4

### LINKED ISSUE #325: Update default concurrency for PX dynos
Hi!

I believe available memory for PX dynos be updated to 14G:

https://github.com/heroku/heroku-buildpack-php/blob/e19cd3062208c1be2ad90130744a048abadc434b/bin/heroku-php-nginx#L301-L306

Reference: https://devcenter.heroku.com/articles/dyno-types

### LINKED ISSUE #374: Configuring memory limit for workers
Currently, I don't think there is a way to configure a different memory limit for web dynos and worker dynos. But this would be a great feature, given we have different needs:

- for web dynos, you want to keep the memory limit low (but still enough for the needs of the application), to allow for concurrency at the PHP-FPM level. The buildpack will properly guess the number of PHP-FPM workers based on the configured memory limit and on the dyno size.
- for worker dynos, the optimal solution would be to have the memory limit equal to (or close to, in case keeping a bit of extra memory for something else makes sense) the dyno memory, as there is only one process running.

For now, I configured my Procfile to use `php -d memory_limit=512M bin/console ...` to override the memory limit of the php.ini when running the worker. But this does not adapt automatically when scaling the dyno to bigger sizes.

An idea I have would be that the buildpack exposes an env variable with the guessed worker memory limit, that could be used in such command (or a script returning it, to use it as ``php -d memory_limit=`vendor/bin/heroku-buildpack-worker-memory-limit` bin/console`` if that's easier).

What do you think about that @dzuelke ?

### LINKED ISSUE #376: PHP 7.4 support
When adding `"php": "^7.4"` to composer.json, the build fails (using `web: vendor/bin/heroku-php-nginx`):

    composer.json/composer.lock requires php ^7.4 -> no matching package found.

dzuelke:
Adds PHP 7.4.2.

Also:
- `$COMPOSER_MEMORY_LIMIT` defaults to dyno memory if information is available
- `$COMPOSER_MIRROR_PATH_REPOS` defaults to 1
- `$COMPOSER_NO_INTERACTION` defaults to 1
- `$COMPOSER_PROCESS_TIMEOUT` defaults to 0 at app runtime
- Build PHP-FPM with `/proc/…/mem` based process tracing
- Log slow PHP-FPM requests after 3 seconds by default for PHP 7.4
- Terminate slow PHP-FPM requests after 30 seconds by default for PHP 7.4
- Refactor `$WEB_CONCURRENCY` calculation to use `/sys/fs/cgroup/memory/memory.limit_in_bytes` if available
- Use all available instance RAM when calculating `$WEB_CONCURRENCY` for PHP 7.4+ running on Performance-L dynos
- Use all available instance RAM as default PHP CLI `memory_limit`

Closes #325 
Closes #374 
Closes #376


laurentvd:
👍

shyim:
Can I increase myself the 30s timeout? 🤔

dzuelke:
What timeout, @shyim? The FPM process termination? The Heroku router will interrupt any request that takes longer than 30 seconds and serve an error page to the client, so it makes no sense to keep a request alive longer than that on the PHP side. If you want to run long-running tasks, use a job queue and worker dynos.

(but yes, you could override it with a custom FPM config)

shyim:
I have an application currently on 7.3. I know I get that timed out page. But the reqeust is still processed. So it's fine for my task. I increased already max_execution_time using .user.ini. But It didn't worked

stof:
@dzuelke is there any easy way to override the slowlog configuration ? I have endpoints which are slower than 3s, and these logs are now spamming my heroku logs since upgrading to PHP 7.4.
I consider them spam, because these requests time are already logged by the heroku router (as a single line of logs rather than many, so more compact) and monitored by my APM tool (which is where I look at them).

PR #389, opened 2020-02-16
Write stdlib to a tmp dir instead of the bpdir 

jkutner:
In Cloud Native Buildpacks, the buildpack dir is read-only. This change fixes a failure on https://github.com/heroku/pack-images/pull/65

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -28,17 +28,17 @@ trap 'err_trap' ERR
 
 # stdlib
 # download to a file first, as the function restarts the entire download on code 18 connection reset (not all servers support -C)
-curl_retry_on_18 --fail --silent --location -o $bp_dir/bin/util/stdlib.sh https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh || {
+curl_retry_on_18 --fail --silent --location -o ${TMPDIR:-/tmp}/stdlib.sh https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh || {
 	error <<-EOF
 		Failed to download standard library for bootstrapping!
-		
+
 		This is most likely a temporary internal error. If the problem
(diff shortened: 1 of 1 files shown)


bin/compile
```diff
@@ -28,17 +28,17 @@ trap 'err_trap' ERR
...
+
 		This is most likely a temporary internal error. If the problem
 		persists, make sure that you are not running a custom or forked
 		version of the Heroku PHP buildpack which may need updating.
 	EOF
 }
-source $bp_dir/bin/util/stdlib.sh
-rm $bp_dir/bin/util/stdlib.sh
+source ${TMPDIR:-/tmp}/bin/util/stdlib.sh
+rm ${TMPDIR:-/tmp}/bin/util/stdlib.sh
```
qw3r: Hi @jkutner,

i think you missed to remove the `/bin/util` path of the path

jkutner: thanks

qw3r:
Another thing: I think there's an other occasion when the script tries to write to the buildpack's directory which you have to change. 
So instead of using `$bp_dir/export` `$tmp_dir/export` would do the trick (given `$tmp_dir` being declared as `tmp_dir=${TEMPDIR:-/tmp}`)

dzuelke:
That won't work, @qw3r, since the `export` script has to be written to the buildpack directory, that's how the V2 API works.

Is the directory marked readonly on the file system, i.e. will a `[[ -w "$bp_dir" ]]` fail?

jkutner:
I think we'll need to back-up and fix this in https://github.com/heroku/cnb-shim. It could copy the target buildpack source code to a tmp dir (which is writeable) instead of running it directly from the bp_dir

jkutner:
The new `heroku/buildpacks:18` has been released, and this should work now. I'm going to close this PR, but please let us know if you continue to have problems and we'll reopen.

PR #393, opened 2020-02-22
Don't create any file inside the buildpack's directory

qw3r:
Fixes a recent change in the buildpack which made the buildpack's directory read-only for heruku user.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -15,6 +15,7 @@ cache_dir=$2/php
 mkdir -p "$cache_dir"
 env_dir=${3:-} # Anvil has none
 bp_dir=$(cd $(dirname $0); cd ..; pwd)
+tmp_dir=${TMPDIR:-/tmp}
 
 export BPLOG_PREFIX="buildpack.php"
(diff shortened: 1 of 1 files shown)


jkutner:
Moving the `export` script out of the buildpack dir will break the old version (v2) of the buildpack. We'll need to add a conditional so that it is only written when not running as a CNB.

qw3r:
Hi @jkutner, 

thank you very much for taking the time and looking into this. I don't know just yet how to check if it's running as CNB, i'll look into it. If you happen to know, please don't hesitate to change my PR ;)

jkutner:
@qw3r thanks for working on this PR, but I think we can fix this more broadly with https://github.com/heroku/cnb-shim/commit/eece77504e9fd3fdf78b029c0bd804373c53d02d

I'll close this PR when the new pack-images builder image is released.

jkutner:
The new `heroku/buildpacks:18` has been released, and this should work now. I'm going to close this PR, but please let us know if you continue to have problems and we'll reopen.

qw3r:
Thanks Joe, fixing the the shim is much much better than my ignorant solution ;)

PR #394, opened 2020-02-28
v172

dzuelke:
The usual version bumps.

A few more system libraries are now used, including the newly available `libonig`, which re-enables mbstring regex on PHP 7.4 for 7.4.3.

Tests are now in place to ensure those essential system libs are used correctly during builds.


Jbithell:
Related stackoverflow question: https://stackoverflow.com/a/60455634/3088158

PR #404, opened 2020-04-24
Support `python` not being symlinked to `python2`

edmorley:
Historically the Ubuntu `python2.7` package has symlinked `python` to `python2`. However starting in Ubuntu 20.04, this symlink is no longer created, and one must explicitly call `python2` (or `python3`) instead.

Unfortunately we cannot just `s/python/python2/` throughout, since this buildpack aims to also support OS X for local development workflows, and on OS X system Python (which is Python 2) is only available as `python` and not also as `python2`. Whilst homebrew Python does have `python{2,3}` binaries, we don't want to require use of it.

As such, for the scripts that may be run locally (`bin/heroku_*`), they must now check for the presence of either `python2` or `python`. 

This change is a no-op for current stacks, but is landing now to reduce the size of the Heroku-20 support PR.

In the future we'll want to convert these usages to Python 3, however we are explicitly not doing so for now, since:
* this work is holding up the Heroku-20 release, and a no-op change is safer to land quickly than changing to start using Python 3 for the first time
* Python 3 is not available out of the box on OS X, so we'd still have to use Python 2 there, which would mean needing to test against multiple Python versions in CI to ensure continued Python 2 compatibility
* all current stacks using the same version of Python 2 (2.7) but for Python 3 their version differs (3.4 on cedar-14, 3.5 on heroku-16, 3.6 on heroku-18 and 3.7 on heroku-20)

Refs [W-7491249](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008NqBgIAK/view).

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -94,7 +94,7 @@ composer_lock_parse_error=$(
 
 # a bunch of sanity checks first
 if [[ -s "$COMPOSER" ]]; then
-	cat "$COMPOSER" | python -mjson.tool &> /dev/null || {
+	cat "$COMPOSER" | python2 -mjson.tool &> /dev/null || {
 		mcount "failures.composer_json.lint"
 		error <<-EOF
 			Basic validation for '$COMPOSER' failed!
@@ -112,7 +112,7 @@ if [[ -s "$COMPOSER" ]]; then
 		EOF
 	}
 	if [[ ! -f "$COMPOSER_LOCK" ]]; then
-		cat "$COMPOSER" | python -c 'import sys, json; sys.exit(bool(json.load(sys.stdin).get("require", {})))' 2> /dev/null || {
+		cat "$COMPOSER" | python2 -c 'import sys, json; sys.exit(bool(json.load(sys.stdin).get("require", {})))' 2> /dev/null || {
 			mcount "failures.composer_lock.missing"
 			error <<-EOF
 				No '$COMPOSER_LOCK' found!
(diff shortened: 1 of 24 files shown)


edmorley:
@dzuelke I don't understand the failure on the Heroku-18 Travis job?
https://travis-ci.org/github/heroku/heroku-buildpack-php/jobs/679146739

The error messages are not that easy to parse; is this just flaky tests? :-)

edmorley:
@dzuelke Hi! Did you have a chance to look into the proposed change of supporting python2+3 discussed on Slack? I'm keen to get Heroku-20 support added soon, and this work is a blocker for that.

bin/compile
```diff
@@ -16,6 +16,10 @@ mkdir -p "$cache_dir"
 env_dir=${3:-} # Anvil has none
 bp_dir=$(cd $(dirname $0); cd ..; pwd)
 
+# Required since OS X's Python 2 binary is named `python` not `python2`,
+# and on newer Ubuntu `python` no longer exists.
+python_cmd=$(command -v python2 > /dev/null && echo python2 || echo python)
```
edmorley, line 21: @dzuelke On Slack using something like `which python2 python` was mentioned, however when given multiple command names, `which` will return the paths to all of them. As such one has to instead use something like:

```bash
python_cmd=$(which python2 || which python)
```

However in the case of neither `python2` nor `python` existing, this will result in `$python_cmd` being the empty string, which will result in confusing error messages when it's used later.

As such we would have to use something like:

```bash
python_cmd=$(which python2 || which python || echo "python")
```

Since we're not really after the path (and only what command name to use) and also `which` varies depending on platform ([see here](https://stackoverflow.com/a/677212)), I instead went for a command+echo approach.

edmorley:
This was cherry picked into #411 as 04c5e145250875259b436fd39212d7c70ed51702.

PR #411, opened 2020-05-25
v175

dzuelke:


rflavien:
With this PR we figured out that our PHP buildpack were running with the python provided by the python buildpack.
Now all our builds are failing with `ImportError: No module named site`. If we remove python from our buildpacks, the builds works but our application can't run without it.
Do you know how we can handle it ?

dzuelke:
Hi @rflavien, thanks for the heads up. Tracking in #412.

dzuelke:
Fixed, @rflavien

rflavien:
Awsome ! Thank you :)

PR #433, opened 2020-09-18
create shared hatchet tests

lillianzhang331:
* Test deploying the getting started guide works
* Test that builds fail when a bad version is specified
* Test that all paths set by the buildpack are absolute instead of relative
* Test CI deploys run tests and use the cache
* Test upgrading stack invalidates the cache
* Test cache for regular deploys is used on repeated deploys


hatchet.lock
```diff
@@ -1 +1,3 @@
---- []
+---
+- - test/fixtures/repos/github/php-getting-started
+  - master
```
schneems, line 3: We should be able to switch this to `main` as long as heroku_hatchet is updated to 7.1.2 or later. Last release is 7.1.4

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      run!('git commit --allow-empty -m "heroku-16 migrate"')
+      app.push!
+      expect(app.output).to include("Loading from cache")
+    end
+  end
+
+  it "should not restore cache if the stack did not change" do
+    Hatchet::Runner.new('php-getting-started', stack: "heroku-16").deploy do |app, heroku|
+      app.update_stack("heroku-16")
+      run!('git commit --allow-empty -m "cedar migrate"')
```
schneems, line 62: If you update to the latest hatchet there's an

```
app.commit!
```

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    app.run_ci do |test_run|
+      expect(test_run.output).to match("mockery/mockery")
+      expect(test_run.output).to include("Downloading")
+      test_run.run_again
+      expect(test_run.output).to include("Loading from cache")
+      expect(test_run.output).to_not include("Downloading")
+    end
+  end
+
+  it "should not restore cached directories" do
```
schneems, line 50:
```suggestion
  it "should not restore cached directories when changing stack" do
```

schneems (commented):
The tests all make sense to me. I think david might want to re-organize the locations, but it's good to have an initial review and either he can ask for them to be moved or move them to where he wants directly.

Thanks a ton for your work here. I want @dzuelke to take an indepth look to see if he's got any questions.

schneems:
There's failed tests, but it doesn't look like it's coming from this work. It looks like problems with already existing tests. @dzuelke know what's going on here?

```
 1) A PHP application on Heroku CI installs dev dependencies

     Failure/Error:

       app.run_ci do |test_run|

       	expect(test_run.output).to match("mockery/mockery")

       end



     RuntimeError:

       Command: git add .; git commit --allow-empty -m next failed: 

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:524:in `local_cmd_exec!'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:329:in `commit!'

     # ./test/spec/spec_helper.rb:78:in `setup!'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:312:in `in_dir_setup!'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:445:in `block in run_ci'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:356:in `block (2 levels) in in_directory'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:354:in `chdir'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:354:in `block in in_directory'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:352:in `in_directory'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/heroku_hatchet-7.2.0/lib/hatchet/app.rb:434:in `run_ci'

     # ./test/spec/ci_spec.rb:6:in `block (2 levels) in <top (required)>'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `loop'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'

     # /home/travis/.rvm/gems/ruby-2.5.3/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

[TEST GROUP 7] 

Finished in 3.81 seconds (files took 1.49 seconds to load)

1 example, 1 failure
```

Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'heroku_hatchet', "~> 7.0"
```
dzuelke, line 3: Please retain that version selector; otherwise, a `bundle update` would bump to Hatchet v8, v9 etc, which is likely to change its API in a breaking fashion

test/spec/spec_helper.rb
```diff
@@ -1,3 +1,5 @@
+ENV['HATCHET_BUILDPACK_BASE'] = 'https://github.com/heroku/heroku-buildpack-php.git'
```
dzuelke, line 1: What is this needed for?

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
+require_relative "spec_helper"
+
+describe "A PHP application" do
+
+  it "works with the getting started guide" do
+    Hatchet::Runner.new("php-getting-started").tap do |app|
```
dzuelke, line 6: Please use `new_app_with_stack_and_platrepo` instead to ensure that the platform repository (which the buildpack pulls PHP etc from) the tests run against is the staging variant - `develop` and other branches then run against that, and `main` and tags run against the stable platform repo. That way, new PHP versions etc are tested before a merge to `main`.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      app.deploy do
+      # deploy works
+      end
+    end
+  end
+
+  it "checks for bad version" do
+    Hatchet::Runner.new("php-getting-started", allow_failure: true).tap do |app|
```
dzuelke, line 14: Same as above with `new_app_with_stack_and_platrepo`.

Also, do not use `allow_failure` here, but instead explicitly assert a failure inside the test body, to ensure that that particular failing behavior happens, and not something completely different.

The purpose of `allow_failure` is to have a test in place that ensures future behavior, but doesn't fail for the moment until e.g. the bug in question is fixed (you'd use that for "known issues" etc).

The problem is that if the `expect()` further down does not succeed (because e.g. it accepts that version given constraint), the test would still pass.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  it "have absolute buildpack paths" do
+    buildpacks = [
+      :default,
+      "https://github.com/sharpstone/force_absolute_paths_buildpack"
+    ]
+    Hatchet::Runner.new("php-getting-started", buildpacks: buildpacks).deploy do |app|
+      #deploy works
```
dzuelke, line 35: Does this test fail if the deploy fails if nothing is asserted?

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+    end
+  end
+
+  it "have absolute buildpack paths" do
+    buildpacks = [
+      :default,
+      "https://github.com/sharpstone/force_absolute_paths_buildpack"
+    ]
+    Hatchet::Runner.new("php-getting-started", buildpacks: buildpacks).deploy do |app|
```
dzuelke, line 34: Same as above with `new_app_with_stack_and_platrepo`.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      expect(test_run.output).to match("mockery/mockery")
+      expect(test_run.output).to include("Downloading")
+      test_run.run_again
+      expect(test_run.output).to include("Loading from cache")
+      expect(test_run.output).to_not include("Downloading")
+    end
+  end
+
+  it "should not restore cached directories when changing stack" do
+    Hatchet::Runner.new("php-getting-started", allow_failure: true, stack: "heroku-18").deploy do |app, heroku|
```
dzuelke, line 51: Same as above with `new_app_with_stack_and_platrepo` and `allow_failure`.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      app.update_stack("heroku-16")
+      app.commit!
+      app.push!
+      expect(app.output).to include("Loading from cache")
+    end
+  end
+
+  it "should not restore cache if the stack did not change" do
```
dzuelke, line 59: I'm not sure what exactly this is testing, given how it's a no-op.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    app.run_ci do |test_run|
+      expect(test_run.output).to match("mockery/mockery")
+      expect(test_run.output).to include("Downloading")
+      test_run.run_again
+      expect(test_run.output).to include("Loading from cache")
+      expect(test_run.output).to_not include("Downloading")
+    end
+  end
+
+  it "should not restore cached directories when changing stack" do
```
dzuelke, line 50: You want this test to fail, right? :)

The test actually succeeds, because the output **does** contain "Loading from cache" (see next paragraph). Since you're using `allow_failure` instead of explicitly asserting that the deploy output does **not** contain "Loading from cache", the test doesn't actually ensure anything - if the `expect()` succeeds, it passes, and if it does not, it passes, too. Same story as with the other `allow_failure` cases further above, except in this case, the problem is actually occurring!

The PHP buildpack does not have a stack-specific cache. Its cache will be re-used between stacks, and that is intentional, because the cache only contains userland PHP code, nothing compiled.

test/spec/test_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    Hatchet::Runner.new("php-getting-started", allow_failure: true, stack: "heroku-18").deploy do |app, heroku|
+      app.update_stack("heroku-16")
+      app.commit!
+      app.push!
+      expect(app.output).to include("Loading from cache")
+    end
+  end
+
+  it "should not restore cache if the stack did not change" do
+    Hatchet::Runner.new('php-getting-started', stack: "heroku-16").deploy do |app, heroku|
```
dzuelke, line 60: `new_app_with_stack_and_platrepo` again please

dzuelke (changes requested):

Gemfile
schneems: I would ask that you not lock. I'll do my best to keep things stable and I'll deprecate before breaking whenever possible. Before starting this shared test work I assumed everyone was on a reasonably recent version of hatchet, but many were several major releases behind.

Your dependencies are still locked in the Gemfile.lock, to update you have to:

```
gem install heroku_hatchet
 && bundle update heroku_hatchet
```

It won't just update itself.

test/spec/spec_helper.rb
schneems: Running tests locally.

test/spec/test_spec.rb
schneems: > do not use allow_failure here, but instead explicitly assert a failure inside the test body, to ensure that that particular failing behavior happens, and not something completely different.

This won't work if you don't pass in `allow_failure` here. By default hatchet will raise an error:

```
remote: Compressing source files... done.
remote: Building source:
remote:
remote: -----> PHP app detected
remote: -----> Bootstrapping...
remote:
remote:
remote:   [UnexpectedValueException]
remote:   Could not parse version constraint 7.badversion: Invalid version string "7.badversion"
remote:
remote:
remote: config [-g|--global] [-e|--editor] [-a|--auth] [--unset] [-l|--list] [-f|--file FILE] [--absolute] [--] [<setting-key>] [<setting-value>]...
remote:
remote:
remote:  !     ERROR: An unknown internal error occurred.
remote:  !
remote:  !     Contact Heroku Support for assistance if this problem persists.
remote:  !
remote:  !     Stack trace follows for debugging purposes:
remote:  !     254 main /tmp/codon/tmp/buildpacks/d91a5b848ec0d2628b25b3bfe70a059132501cef/bin/compile
remote:
remote:  !     Push rejected, failed to compile PHP app.
remote:
remote:  !     Push failed
remote: Verifying deploy...
remote:
remote: !	Push rejected to hatchet-t-79874d6deb.
remote:
To https://git.heroku.com/hatchet-t-79874d6deb.git
 ! [remote rejected] HEAD -> main (pre-receive hook declined)
error: failed to push some refs to 'https://git.heroku.com/hatchet-t-79874d6deb.git'
Destroying "hatchet-t-79874d6deb": f2790ef2-3566-4acb-9b02-b43d3425d839, total_app_count: 274, hatchet_app_count: 20/20, finished: 0, unfinished: 20
```
 
> The problem is that if the expect() further down does not succeed (because e.g. it accepts that version given constraint), the test would still pass.

If accepts the version constraint then it would not output "Could not parse version constraint 7.badversion", no?

test/spec/test_spec.rb
schneems: Correct. An exception is raised if the deploy fails by default. Here the logic and checks are actually being performed in the buildpack:

https://github.com/sharpstone/force_absolute_paths_buildpack

See the readme for more info

Gemfile
edmorley: Also if one locks in `Gemfile`, I'm presuming Dependabot might not open a PR to update it, which defeats the point :-)

test/spec/test_spec.rb
schneems: > instead of explicitly asserting that the deploy output does not contain "Loading from cache"

Then the solution is to add a

```
      expect(app.output).to_not include("Loading from cache")
```

Before the stack update and deploy

> The PHP buildpack does not have a stack-specific cache. Its cache will be re-used between stacks, and that is intentional because the cache only contains userland PHP code, nothing compiled.

Right, so I think we want to keep the test and assert that the stack doesn't accidentally get cleared on stack upgrade.

test/spec/test_spec.rb
schneems: Good catch. I'm guessing some of these were copy/pasta

test/spec/test_spec.rb
schneems: Agreed. I think we should remove it

test/spec/test_spec.rb
schneems: 👍

Gemfile
schneems: > Also if one locks in Gemfile, I'm presuming Dependabot might not open a PR to update it, which defeats the point :-)


Exactly. Dependabot won't send you a PR if you've told bundler to only resolve within this range

test/spec/test_spec.rb
schneems: There was a misunderstanding over exactly what `allow_failure` does and doesn't do. We resolved it in slack

schneems:
Tests are fine, but super flappy. They keep hanging. Travis borked my permissions so I can't individually restart a test.

dzuelke (approved):

PR #447, opened 2020-11-20
Composer 2 support

### LINKED ISSUE #440: Support for Composer 2
Now that Composer 2 has been [released](https://blog.packagist.com/composer-2-0-is-now-available/), can we have it supported by the buildpack? Currently, `composer.lock` files generated by Composer 2 simply do not work with Heroku due to the Composer version being hardcoded.

dzuelke:
- platform install still happens with Composer 1 for now
- Composer is now a `heroku-sys-program` (new type) and installed via the platform install step
- it uses plugin API `provide`s to auto pick the right version based on lock file
- only either v1 or v2 are installed, not both (`composer1` is used only for platform installs and then removed)
- no more download progress output since it messed up output massively (see commit)
- also fixed a New Relic info log output regression

Closes #440


edmorley (approved):
Looks good to me (so much as I can understand it without having ever used Composer before).
Travis is failing for Cedar-14 but presume since binaries weren't generated since EOL etc.

PR #469, opened 2021-04-12
Move to CircleCI

dzuelke:


edmorley (approved):
Awesome! 

As an added bonus - the end to end times look a lot quicker than Travis CI (5-6mins vs 8-9 mins, and that's not even including shorter queue times).

PR #479, opened 2021-06-01
Remove heroku-16

dzuelke:
Drops all heroku-16 references, drops all references to PHP before 7.1 (mostly in build formulae), shifts extension build formulae base from 5.6 to 7.1.


edmorley (approved):
🎉

PR #488, opened 2021-06-25
Blackfire: tests, agent v2, separate agent package, blackfireio/integration-heroku buildpack support

dzuelke:
Until now, the `ext-blackfire` extension also always bundled the `blackfire-agent` program. That program relays telemetry from a language runtime to the Blackfire servers in the background.

This PR:
- splits the extension and agent into two packages, so they can be versioned independently
- bumps the agent to v2 (now just called "blackfire")
- skips installation of "blackfire" if a user has https://github.com/blackfireio/integration-heroku set as a buildpack on their app - in that case, it uses the version that buildpack installed
- adds tests for all of this behavior


schneems (approved):

schneems (approved):

lolautruche (commented):
Would you have the possibility to automatically download the latest version of Blackfire extension of needed?

dzuelke:
We can't do that for compliance reasons, unfortunately - released versions must be fixed and mirrored by us for audit log purposes.

Longer-term, I guess the best solution is for us to simply not provide the CLI at all, but it'd be great if the user experience was reasonable in that case (i.e., meaningful error message instructing folks to install your buildpack).

Of course, another solution would be for you to also provide the extension packages ;) We can tackle that after I migrated the platform installer plugin to Composer v2.

lolautruche:
OK understood.

PR #498, opened 2021-09-21
More platform installer tests

dzuelke:
- covers extension ordering in generated repo explicitly (see also https://github.com/composer/composer/pull/10110)
- test forward compat with packages that are built for our upcoming Composer 2 installer
- test userland packages providing native extensions edge cases

GUS-W-8105408
GUS-W-9889098

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,9 +23,16 @@ jobs:
           command: |
             gem install bundler --version="~> 1"
       - run:
-          name: Install PHP
+          name: Update apt
           command: |
             sudo apt-get update
+      - run:
+          name: Install Python
+          command: |
+            sudo apt-get install python
+      - run:
+          name: Install PHP
+          command: |
             sudo apt-get install php-cli
       - run:
           name: Install Composer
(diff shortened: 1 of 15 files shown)


dzuelke:
(tests are currently failing due to a `/dev/std{err,out}` permissions regression in CR with `heroku run`)

.circleci/config.yml
```diff
@@ -23,9 +23,16 @@ jobs:
...
             gem install bundler --version="~> 1"
       - run:
-          name: Install PHP
+          name: Update apt
           command: |
             sudo apt-get update
+      - run:
+          name: Install Python
+          command: |
+            sudo apt-get install python
```
edmorley, line 32: Since these are single commands, the multi-line syntax can be removed (the `|`). Also I wonder if you might be able to get away with `--no-install-recommends` to speed up CI slightly.

edmorley (approved):

.circleci/config.yml
dzuelke: fcd2482 (and no recommended installs, so no benefit in this case)

PR #499, opened 2021-09-28
v198

dzuelke:
- work around CR/LXC2 `/dev/std{out,err}` permissions regression in New Relic logging (affected CI; GUS-W-9958053)
- work around CR/LXC2 `/dev/std{out,err}` permissions regression in `waitforit.sh` test utility (affected CI; GUS-W-9958077)
- wait for `blackfire` agent startup on boot in the same way we're already doing  it for New Relic (caused tests to sometimes be flaky if `blackfire` didn't start quickly enough)
- use a forked version of Blackfire's "integration buildpack" (that can also be used instead of our own package) that does the same waiting (caused tests to sometimes be flaky if `blackfire` didn't start quickly enough)
- fix tests around caching that were using php-getting started (which now installs using Composer2, so output changed)
- routine runtime/extension/package updates (GUS-W-9831719)


dzuelke:
(yes, this all needs to be in one PR, since most commits fix tests that broke due to external changes, and some binary packages needed rebuilding for these changes to take effect)

edmorley (approved):

doriskwan (approved):

PR #527, opened 2022-01-03
v204 with Composer 2.2 support and install forward compat

### LINKED ISSUE #526: Platform package install fails with Composer 2.2 lock file
Several people, including me, are experiencing an issue similar to #481 using the latest version of Composer. As far as I understand, it would be enough to bump the Composer version of this buildpack. I was able to work around the issue by setting the `plugin-api-version` to `2.1.0`.

```
!     > You are using Composer 1 which is deprecated. You should upgrade to Composer 2, see https://blog.packagist.com/deprecating-composer-1-support/
!     > Loading repositories with available runtimes and extensions
!     > Updating dependencies
!     > Your requirements could not be resolved to an installable set of packages.
!     >
!     >   Problem 1
!     >     - The requested package composer-plugin-api could not be found in any version, there may be a typo in the package name.
```

I am quite new to this, so did I understand correctly that future updates of Composer (which are automatically installed on my machine using homebrew) would break my deployments again as it happens today? If so, could/should that be fixed by improving the Composer version detection logic (I do not understand completely how that works yet)?

dzuelke:
Update Composer to 2.2.3, and allow any e.g. 2.x.y lock file to be installed by any 2.x.y version, as installs are forward compatible (https://twitter.com/seldaek/status/1474477031329050625).

Closes #526, too.

GUS-W-10372313
GUS-W-10372327


test/spec/composer_spec.rb
```diff
@@ -25,6 +25,14 @@
 			end
 		end
 	end
+	context "with a composer.lock generatead by a future version 2 of Composer" do
```
edmorley, line 28:
```suggestion
	context "with a composer.lock generated by a future version 2 of Composer" do
```

edmorley (approved):

PR #537, opened 2022-02-10
Dedicated packages for PHP shared extensions

dzuelke:
Shared extensions bundled with PHP have special handling in the installer. The reason is that if a `composer.json` `require`s "`ext-bcmath`" (bundled with PHP), Composer will not fire a separate install event, as the previous installation of package "`php`" already satisfies the requirement. So the installer plugin maintains a list of all extension requirements, and, when installing a PHP package, manually enables these extensions by writing an INI config for PHP to load the shared module, just the same way it would for a separate "external" extension package.

Aside from adding complexity to the code (which isn't removed yet, as it would be a BC break for people using third-party repositories, and because only our maintained PHP versions get rebuilds with this change), this leads to an inconsistency with how so-called "polyfills" are treated.

Polyfills are userland packages that claim to `provide` a native extension, by implementing the functionality in userland. Popular examples are e.g. https://packagist.org/packages/symfony/polyfill-mbstring or https://github.com/phpseclib/mcrypt_compat for extensions that are (or at some point were) bundled with PHP; https://packagist.org/packages/symfony/polyfill-uuid or https://packagist.org/packages/alcaeus/mongo-php-adapter are polyfills for third-party extensions.

When these are present, the native extension, if it were a separate package, would not be selected for installation by Composer's dependency solver. However, our installer plugin can't know that - it only knows that somewhere in `composer.json`, "`ext-mbstring`" is `require`d, and that package "`php`" says it `replace`s that package, so it gets enabled. It can't know that "`symfony/polyfill`" claims to `provide` a suitable version of "`ext-mbstring`".

That behavior does not lead to problems, as the polyfills always contain logic to fall back to the native extension they're "emulating" in userspace if it's present, but it is inconsistent with how the installer behaves for extensions that are "real" packages, e.g. "`ext-uuid`", which will not be installed or enabled when https://packagist.org/packages/symfony/polyfill-uuid is used.

This change causes the repository generator to expand bundled shared PHP extensions into their own packages, with a `dist` `type` that will cause the installer to not download anything. The packages will be treated like any other extension by Composer for dependency resolution, and will go through a normal installer event inside Composer, meaning the same logic for enabling extensions as for third-party extensions is used.

In a future version of the installer, the legacy logic for handling shared bundled PHP extensions can then be removed.

GUS-W-9889113


dzuelke:
Also contains a change in preparation of GUS-W-9889127 by declaring `.native` variants of all extensions as `replace`d by the appropriate packages.

support/build/_util/include/manifest.py
```diff
@@ -23,7 +27,36 @@
 	"replace": json.loads(sys.argv[7]) if len(sys.argv) > 7 else {},
 	"provide": json.loads(sys.argv[8]) if len(sys.argv) > 8 else {},
 	"extra": json.loads(sys.argv[9]) if len(sys.argv) > 9 else {},
-	"time": datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S")
+	"time": os.getenv("NOW", datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S"))
 }
 
+# if it's a PHP manifest, we will generate full manifests for each shared extension into extra.shared; mkrepo.sh will then expand those into actual package declarations when it generates the repo
```
edmorley, line 33: Could you wrap these comments? I have to scroll sideways in the web UI to read them when reviewing otherwise, which makes reviews slower.

edmorley (approved):

dzuelke:
Added a commit that causes the tests to run against the stable bucket (and skips one related test for now); this will be merged to `main`, tagged, and released through buildpack registry without syncing the develop bucket to stable.

Then a follow-up PR will restore the skipped test; once approved, the develop bucket is synced to stable, then the PR gets merged to `main`, tagged, and released through buildpack registry.

dzuelke:
Interesting. Thanks for bringing this up so quickly, @coffe4u!

I am guessing you're also using Symfony? That likely means you also have `symfony/mbstring-polyfill`, which now prevents the installation of the native `ext-mbstring` because Symfony never selects the latter as a candidate during solving (since `symfony/mbstring-polyfill` declares a `provide` of `ext-mbstring`).

Not to worry, a follow-up PR that addresses exactly this is already in flight, although it was intended as a performance optimization, and not as a workaround for this case (PHPUnit must be checking `extension_loaded()` for mbstring).

In the meantime, you could pin https://packagist.org/packages/symfony/polyfill-mbstring to `~1.23.1`; the `provide` for `ext-mbstring` was only introduced recently (in 1.24.0).

coffe4u:
@dzuelke thank you for the quick response!

I'm using Laravel which in turn uses Symfony. I ran `composer require symfony/polyfill-mbstring:~1.23.1` and now things are working again. I appreciate not only the suggestion on how to move forward for now but also the information on why this happened. Thanks again!

dzuelke:
This is now implemented in branch `composer2installer`:

```
remote: -----> Installing platform packages...
remote:        - php (8.1.2)
remote:        - apache (2.4.52)
remote:        - composer (2.2.5)
remote:        - nginx (1.20.2)
remote:        NOTICE: detected userland polyfill packages for PHP extensions
remote:        NOTICE: now attempting to install native extension packages
remote:        Installing extensions provided by symfony/polyfill-mbstring:
remote:        - ext-mbstring (bundled with php)
```

Please give it a try if you can:

```
heroku buildpacks:set https://github.com/heroku/heroku-buildpack-php.git#composer2installer
```

and report back any issues!

coffe4u:
@dzuelke Dependabot created a `Bump symfony/polyfill-mbstring from 1.23.1 to 1.24.0` PR for me this morning so on that branch, I updated my `app.json` config and replaced the testing PHP buildpack with `https://github.com/heroku/heroku-buildpack-php.git#composer2installer`. PHPUnit ran successfully and everything appears to work as expected. Here is the relevant build output:

```
-----> Building on the Heroku-20 stack
-----> Fetching https://github.com/heroku/heroku-buildpack-php.git#composer2installer buildpack...
       buildpack downloaded
-----> Fetching heroku/nodejs buildpack...
       buildpack downloaded
-----> Fetching stebogit/heroku-mysql-cli buildpack...
       buildpack downloaded
-----> Fetching https://github.com/heroku/heroku-buildpack-google-chrome buildpack...
       buildpack downloaded
-----> PHP app detected
-----> Bootstrapping...
-----> Preparing platform package installation...
-----> Installing platform packages...
       - php (8.0.15)
       - ext-ftp (bundled with php)
       - apache (2.4.52)
       - composer (2.2.5)
       - nginx (1.20.2)
       - ext-sodium (bundled with php)
       - ext-gd (bundled with php)
       NOTICE: detected userland polyfill packages for PHP extensions
       NOTICE: now attempting to install native extension packages
       Installing extensions provided by symfony/polyfill-mbstring:
       - ext-mbstring... 
       Installing extensions provided by symfony/polyfill-iconv:
       - ext-iconv... 
       Installing extensions provided by symfony/polyfill-ctype:
       - ext-ctype... 
```

dzuelke:
Fixed and released (#541) to the official stable buildpack channel.

@coffe4u and everyone else, you can now revert your workarounds.

Thank you for your patience and enjoy your weekend!

PR #542, opened 2022-02-12
Adding Updating CODEOWNERS with ECCN as per Export Control Compliance

svc-scm:


edmorley:
These ECCN PRs are redundant, given that Open Source repositories are exempt from needing an export classification, and the GitHub repository metadata clearly states this repo is `public`, so there is no need for a separate file. In addition, PRs should have a sensible description/commit message (particularly when scripted/opened on mass), and this does not.

PR #549, opened 2022-03-16
Adding start timeout parameter

charlyhue:
php-fpm take more than 2500 ms to start when Datadog appsec is enabled. This change add an optional parameter to configure this timeout.


dzuelke (changes requested):
Can you please change the `timeout` default value to 2500, do an arithmetic operation to divide by 100, and adjust the help output to say `-t <timeout> Startup timeout for launched processes (in ms)`, also mentioning the default value? Providing the timeout as a count is pretty unintuitive.

dzuelke:
Bump?

PR #575, opened 2022-06-13
Switch to the recommended regional S3 domain

edmorley:
Whilst the global S3 endpoint (`s3.amazonaws.com`) still works, AWS now recommends using the appropriate regional endpoint to access the bucket:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#s3-legacy-endpoints

Our buildpack buckets are in `us-east-1`, whose regional domain is `*.s3.us-east-1.amazonaws.com`:
https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_region

Ideally the `S3_REGION` env var would be set to values like `us-east-1` rather than `s3.us-east-1`, however in order that this not be a breaking change for users using custom buckets, I've left the `s3.` prefix in for now. In the future the env var handling could be adjusted to strip the prefix.

GUS-W-11283397.


support/build/README.md
```diff
@@ -169,7 +169,7 @@ The following environment variables are highly recommended (see section *Underst
 
 The following environment variables are optional:
 
-- `S3_REGION`, to be set to the AWS region name (e.g. "`s3-eu-west-1`") for any non-standard region, otherwise "`s3`" (for region "us-east-1")
+- `S3_REGION`, to be set to the AWS region name (e.g. "`s3.eu-west-1`") for any non-standard region, otherwise it will default to "`s3.us-east-1`".
```
edmorley, line 172: The hyphenated form is also considered legacy:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#s3-legacy-endpoints

dzuelke (approved):

PR #579, opened 2022-07-01
v221: Composer hotfixes

dzuelke:
Composer has started failing for `allow-plugins: true` in configs on July 1: https://github.com/composer/composer/issues/10907 / https://github.com/composer/composer/pull/10909


schneems (approved):
LGTM. Thanks!

PR #604, opened 2022-11-10
Use GitHub Actions for CI

dzuelke:
GUS-W-11832719


edmorley (commented):
The old Circle CI config (`.circleci/config.yml`) needs deleting.

Also the following need updating:
- the README badge 
- the ignore section in `buildpack.toml`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
+name: CI
+
+on:
```
edmorley, line 3: I think you'll need a `tags` entry here, like:
https://github.com/heroku/stack-images/blob/c151f87271e02aa7d8524cb46740e6a94355f771/.github/workflows/ci.yml#L8-L9

edmorley (approved):
Thank you!

PR #605, opened 2022-11-16
Change curl timeout for low speed connection

saigontech:
for the low-speed connection environment, the current 5s timeout value seems too low.
I often get `Failed to download standard library for bootstrapping!` error when deploying the app to my homelab server
can we increase it to ~ 60s to make it works more friendly?

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -28,7 +28,7 @@ trap 'err_trap' ERR
 
 # stdlib
 # download to a file first, as the function restarts the entire download on code 18 connection reset (not all servers support -C)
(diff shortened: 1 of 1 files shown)


edmorley (commented):
(switching the code review request from the team alias to the PHP language owner)

dzuelke:
Idea: should we support `CURLOPT_TIMEOUT` and `CURLOPT_CONNECTTIMEOUT` as env vars, and use them there? (these are not standard env vars that curl understands by default, but the names would correspond to the options of `libcurl`)

edmorley:
Fwiw for Python, bumping the fixed timeout very slightly was enough to make the "underpowered server" use-case work without needing to support customising the timeout:
https://github.com/heroku/heroku-buildpack-python/pull/1439

PR #612, opened 2023-01-25
Update Hatchet to latest timeouts-etc branch

dzuelke:
This buildpack uses that branch (https://github.com/heroku/hatchet/compare/main...timeouts-etc) for its timeout handling capability. Its changes have now been ported to include recent Hatchet changes such as the fix for scaling to "basic" instead of "free" after multi-dyno runs, restoring the ability to run CI without bogus failures.

GUS-W-12421825

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,10 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


Malax (approved):

dzuelke:
(the change to build all branches is intentional, too)

PR #633, opened 2023-05-08
merge with heroku

kinsflow:


dzuelke:
Sure! https://devcenter.heroku.com/articles/php-support#customizing-settings-php mentions the simplest way very briefly. It links to https://devcenter.heroku.com/articles/custom-php-settings which documents all ways of customizing `php.ini` settings, and for your particular case of adjusting the memory limit, you should also read https://devcenter.heroku.com/articles/php-concurrency carefully, because changing the memory limit will automatically adjust the number of PHP-FPM worker processes that process your requests.

I assume you don't need 256 MB of RAM for every request, but really a lot less than that, except in one or two cases, so `ini_set()`ting a higher limit for those cases is probably the better approach, as explained in https://devcenter.heroku.com/articles/php-concurrency#runtime-changes-of-memory_limit (check the note in the blue box).

Either way, for globally changing the memory limit, drop a `.user.ini` file with `memory_limit = …` into your web root, and PHP-FPM will pick that up. No need for a config file change etc. It's the approach that https://devcenter.heroku.com/articles/php-concurrency#setting-memory-limit-via-user-ini describes.

kinsflow:
@dzuelke Still struggling with this. I am using dokku and my project is a Laravel project. however, all efforts through .user.ini or fpm_custom.conf ain't working. upload_max_filesize is still the default 2M. Is thee any help possible?

PR #646, opened 2023-07-18
Platform package metadata fixes

dzuelke:
- small typo fix in a printed notice ("`plugin-api-version`", not "`platform-api-version`");
- expose `composer-runtime-api` virtual package through the `composer` package, similar to `composer-platform-api`;
- clean up handling of virtual `.native` package variant for extension packages
  - formulae no longer have to be aware of how this is implemented under the hood;
  - extensions that require another extension now automatically also require the `.native` variant;
  - same for `replace`s;
  - some tests related to all the above

GUS-W-13772872
GUS-W-13772974


edmorley (changes requested):
This seems to be three unrelated changes that have been combined into one PR. 

Could you split it into three PRs? (I realise there are separate commits, but I still think it would be better to get in the habit of preferring separate PRs, since it reduces the cognitive burden on the reviewer, makes it easier to release/revert etc individually if required, plus makes it easier for end users to follow along.)

edmorley (approved):
Approving for now, since I've been promised future PRs will be more split up in the future :-)

PR #650, opened 2023-08-16
Drop lang-common S3 bucket dependency

dzuelke:
This imports the relevant (to this buildpack) telemetry helper functions from https://lang-common.s3.us-east-1.amazonaws.com/buildpack-stdlib/v8/stdlib.sh

GUS-W-13958512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # heroku-buildpack-php CHANGELOG
 
+## v238 (2023-08-??)
+
+### CHG
+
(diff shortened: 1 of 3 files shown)


edmorley (approved):
I suppose all of our classic buildpack's telemetry functions don't do anything any more, since cytokine dropped Librato support as part of Librato sunset, so cytokine no longer sets `BUILDPACK_LOG_FILE`, meaning these all get sent to `/dev/null` at present:
https://github.com/heroku/heroku-buildpack-php/blob/bd5bdce2f7959581bed352d94d7bde875fc51eb0/bin/compile#L20

PR #656, opened 2023-09-28
2023-09 housekeeping

dzuelke:
Drop 86df1f820d3460591347b3dd6eee168080274c48 since it was fixed by https://github.com/heroku/cli/pull/2467.

Update a few dependencies for Bundler and Composer, all together (and have dependabot group these in the future).


edmorley (approved):

dzuelke:
Updated `waitforit.sh` to work around an issue with Focal based instances that would not allow trapping of `SIGPIPE`, which the script needed to operate (tests were hanging forever as a result, because some test dynos would never terminate).

PR #670, opened 2023-12-07
PHP 8.3 preps: test updates and default version range limits

dzuelke:
Update some tests for forward compatibility with 8.3+

Also generally confine the default PHP version to something known; this prevents test failures in unrelated branches running against dev buckets with newer major builds (n.b. yes, we'll eventually re-structure this to allow more than one "develop" bucket).

GUS-W-14454295


Malax (approved):
LGTM

PR #692, opened 2024-02-08
Fix FPM request_terminate_timeout not working

dzuelke:
With `HEROKU_PHP_GRACEFUL_SIGTERM=1`, FPM children ignore SIGTERM (as sent by e.g. Heroku's process supervisor to all processes) so that the boot scripts can perform a clean shutdown by issuing SIGQUIT.

However, FPM internally depends on SIGTERM for terminating requests that exceed the timeout, so that's now being ignored.

We can, however, use SIGINT in this case, as it serves the same purpose.

Tests adapted to cover this case (the termination did previously work in some cases, e.g. when the script simply does a `sleep()`, as that is still interrupted by an incoming signal, even if the signal is being ignored).

GUS-W-14850175


dzuelke:
Builds:
- https://github.com/heroku/heroku-buildpack-php/actions/runs/7753911895 (heroku-20, PHP 8)
- https://github.com/heroku/heroku-buildpack-php/actions/runs/7757405901 (heroku-20, PHP 7.4)
- https://github.com/heroku/heroku-buildpack-php/actions/runs/7753915148 (heroku-22)

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Unreleased]
+## [v245] - 2024-02-08
```
edmorley, line 3: This is an incomplete change for performing a release. The links in the footer need updating too :-) Plus the `[Unreleased]` section should be left in place for the next cycle. These are all things the prepare release automation would do for you, reducing the chance of user error.

Given there are multiple changes already on `main` for this release (and so this PR isn't the only change in the release), it seems cleaner to have the "prepare release" changes be their own PR. (In the scenario where you have no changes on `main` and have a combination PR from `develop`, then I could see the argument for bundling in the release preparation into the combo PR.) 

With the new automation, the creation of the release preparation PR is very quick and can be self-reviewed, meaning there isn't another delay for a review cycle.

Lastly, should this release get delayed after this PR is merged, then with the current PR it would require a manual fixup of the release date (that would require another team member to review), whereas with the automation PR, you can re-trigger the workflow and it will update the "prepare release" PR with the new date.

edmorley: Would you mind omitting this line from the PR, and trying out the release automation at least once? :-)

dzuelke: Alright, @edmorley

edmorley (approved):

dzuelke:
Package syncs: https://github.com/heroku/heroku-buildpack-php/actions/runs/7842181503

PR #715, opened 2024-05-22
Python 3.12 compatibility for build tooling

dzuelke:
Ubuntu 24.04 ships with Python 3.12, which deprecates a few things we use (`distutils.version`, `datetime.datetime.utc*`).

More crucially, [bob-builder](https://github.com/heroku-python/bob-builder) uses `boto`, which due to a weird vendoring of `six` is incompatible with Python 3.12, so we need to update it to [0.0.20, which uses `boto3`](https://github.com/heroku-python/bob-builder/pull/53).

That version picks up `S3_REGION` from the environment, but our `S3_REGION` values so far contained a prefix of "`s3.`".

GUS-W-157948523


support/build/_docker/heroku-20.Dockerfile
```diff
@@ -5,14 +5,19 @@ ENV WORKSPACE_DIR=/app/support/build
...
 ENV STACK=heroku-20
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y python3-pip
+RUN apt-get update && apt-get install -y python3-pip python3-venv
+
+ENV VIRTUAL_ENV=/app/.bob-venv
+RUN python3 -m venv "$VIRTUAL_ENV"
+
+ENV PATH="$VIRTUAL_ENV/bin:/app/usr/bin:$PATH"
```
edmorley, line 17: (optional) The PATH is also updated on line 5 - worth combining the two?

support/build/_docker/heroku-22.Dockerfile
```diff
@@ -5,14 +5,19 @@ ENV WORKSPACE_DIR=/app/support/build
 ENV PATH=/app/support/build/_util:$PATH
 ENV S3_BUCKET=lang-php
 ENV S3_PREFIX=dist-heroku-22-develop/
-ENV S3_REGION=s3.us-east-1
+ENV S3_REGION=us-east-1
 ENV STACK=heroku-22
```
edmorley, line 9: (optional, and potentially for another PR anyway) Docker build args could reduce the duplicate of Dockerfiles here - ie: single `Dockerfile` that takes a `STACK_VERSION` arg, like:
https://github.com/heroku/heroku-buildpack-python/blob/ce39d0cc33c29c2583ff52fe2536e48908b9a0b8/builds/Dockerfile#L1-L5

edmorley (approved):

requirements.txt
```diff
@@ -1,2 +1,3 @@
-bob-builder>=0.0.15
+bob-builder>=0.0.20
 s3cmd>=1.6.0
+natsort
```
edmorley, line 3: CI is failing with `ModuleNotFoundError: No module named 'natsort'`

support/build/_docker/heroku-22.Dockerfile
dzuelke: Yup, these are great, improvement for a future iteration :)

PR #732, opened 2024-06-20
Support cgroups v1 and v2 for memory and CPU limits

### LINKED ISSUE #699: cgroupsv2 support
The buildpack currently reads available RAM from a hard-coded location: `/sys/fs/cgroup/memory/memory.limit_in_bytes`.

This is a cgroups v1 path; cgroups v2 use a set of different soft and hard limits exposed via `/sys/fs/cgroup/mem.{min,low,high,max}`.

For both v1 and v2, groups can be arranged hierarchically, so it's actually necessary to determine the group that applies to the currently running process by parsing the `cgroup` table for the current process in `/proc`.

For cgroups v1, an entry may exist for a particular controller, e.g. "memory". For v2, only a single unified hierarchy exists.

For cgroups v2, all leaf nodes always expose all effective limits for active controllers, meaning no traversal up the tree is necessary to determine a limit. For v1, such path traversal must be performed, as a child group may inherit a limit from its parent.

dzuelke:
Try to find group membership, controller, and mount info from `procfs`, then read a suitable value:
- `memory.limit_in_bytes` for v1
- `memory.high` for v2 (the recommended way of setting limits)
  - if that does not exist, read `memory.max` (e.g. Docker's  `-m`)
    - if that does not exist, read `memory.low` (e.g. Docker's `--memory-reservation`)

The routines guard against silly values (such as Docker's `memory.limit_in_bytes` at `9223372036854771712` on unlimited containers, that would be about 8 Exbibytes).

If nothing is found, the boot scripts try falling back to plainly reading `/sys/fs/cgroup/memory/memory.limit_in_bytes` directly, as before.

Tests cover `set -e` and `set -u` environments, as well as regular and verbose mode of the functions, in all possible permutations.

Closes #699

GUS-W-16050052
GUS-W-16050059

---

Also now uses `nproc` to get the number of cores, falling back to `getconf` for e.g. BSD systems (the boot scripts can be used on their own).

---

For easier review, the test fixtures are in a separate commit.


bin/heroku-php-apache2
```diff
@@ -387,21 +389,43 @@ echo -e "\n[global]\nlog_level = notice" >> "$fpm_config_tmp"
...
-if [[ -f "$mlib" ]]; then
-	[[ $verbose ]] && echo "Reading available RAM from '$mlib'" >&2
-	ram="$(cat "$mlib")"
-else
-	[[ $verbose ]] && echo "No '$mlib' with RAM info found" >&2
-	ram="512M"
-	echo "Assuming RAM to be ${ram} Bytes" >&2
+ram=
+# a hard maximum we never want to exceed, e.g. in case of silly values
+memory_maximum=$(( 8*1024*1024*1024*1024 ))
```
schneems, line 394: I would like a unit here. Would be good to either have it in the variable name or a comment or extracted as a variable:

```suggestion
terabyte=$((1024*1024*1024*1024))
memory_maximum=$((8*terabyte))
```

This is assuming the input `-m` takes bytes, and is supposed to be 8 TB

PR #767, opened 2024-11-22
PHP 8.4

dzuelke:
No blackfire or newrelic extension builds yet with support for 8.4. The phalcon extension is incompatible with 8.4 (they always get new releases some time later), and pcov needs a fix and release first: https://github.com/krakjoe/pcov/pull/111. The imap extension has moved to PECL.

The heroku-20 stack is not getting this version because the stack is already deprecated.

GUS-W-17226361


bin/compile
```diff
@@ -368,9 +368,9 @@ composer validate --no-plugins --no-check-publish --no-check-all --quiet "$COMPO
 status "Preparing platform package installation..."
 
 if [[ $STACK == "heroku-20" ]]; then
-	HEROKU_PHP_DEFAULT_RUNTIME_VERSION="^7.3.0 | ^8.0.0 <8.4"
+	HEROKU_PHP_DEFAULT_RUNTIME_VERSION="^7.3.0 | ^8.0.0 <8.5"
 else
-	HEROKU_PHP_DEFAULT_RUNTIME_VERSION="^8.0.0 <8.4"
+	HEROKU_PHP_DEFAULT_RUNTIME_VERSION="^8.0.0 <8.5"
```
edmorley, line 373: If I'm reading this correctly, apps with an unspecified PHP version can now get PHP 8.4 for the first time. However, this isn't mentioned in the changelog? (The `PHP/8.4.1` entry isn't sufficient to infer that IMO - particularly since  it's a `.1` release, and not `8.4.0`, so doesn't make it clear that this is the first new 8.4 release)

edmorley (approved):

dzuelke:
(n.b. waiting for merge/release of https://github.com/composer/composer/pull/12217, and the Black Friday weekend to tide over)

bin/compile
dzuelke: It's documented that we always use the latest possible version as the default: https://devcenter.heroku.com/articles/heroku-php-behavior#default-runtime

dzuelke:
Builds:
- `heroku-20`:
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/12372455407 (for new `ext-pcov`, `ext-blackfire` and `composer` versions required for PHP 8.4 on other stacks)
- `heroku-22`:
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11980430238
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973020186
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/12372333296
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973146864
- `heroku-24-amd64`:
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11980432039
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973023268
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973160153
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/12372337572
- `heroku-24-arm64`:
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11980433065
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973026059
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/11973162560
  - https://github.com/heroku/heroku-buildpack-php/actions/runs/12372340459

Packages sync dry-run: https://github.com/heroku/heroku-buildpack-php/actions/runs/12396463838

bin/compile
edmorley: I still think it warrants a mention in the changelog, if the buildpack release has the effect of changing the result of a version resolution to a new major version?

edmorley: If only because of:

> particularly since it's a .1 release, and not 8.4.0, so doesn't make it clear that this is the first new 8.4 release

dzuelke: Ah, maybe to clarify - this default only applies to users with basically an empty (or no) `composer.json`. Once you have any requirements for `php` whatsoever in your lock file (and 99.9% of userland libraries specify some kind of version constraint, say `>=7.2.0`), that gets used instead.

The Dev Center changelog entry will mention it, so that's enough IMO (and consistent with previous changelog entries where we did not mention this either).

The constraints there for the default version are more for internal technical reasons anyway (to ensure other branches do not pick up new versions as a default), due to how there is one "global" staging/dev bucket for binaries, rather than something per-branch (moving to which is the longer-term goal).

dzuelke:
Sync: https://github.com/heroku/heroku-buildpack-php/actions/runs/12416682210

PR #794, opened 2025-05-05
Drop support for heroku-20

dzuelke:
This also allows the removal of PHP 7.3, 7.4, 8.0, Composer v1, and a bunch of cruft for older extension versions that are now also gone.

GUS-W-14707054


schneems (dismissed):
- There are 7.3, 7.4, and 8.0 files in the `./conf` directory, please remove those.
- There are PHP 7.4 and 8.0 test fixtures like `./test/fixtures/platform/builder/mkrepo/create-native-replace-require-entries/php-7.4.22.composer.json` please remove or update those

Please run this command and ensure all heroku-20 references are gone or should still be there:

```
$ grep --exclude-dir=".git" --exclude="*/CHANGELOG.md" --exclude="*/shunit2" --exclude="*/composer.lock" --exclude="*/README.md" --exclude="*/no-debug-non-zts-*" -R "heroku-20"
```

dzuelke:
Good catch on the configs, thanks.

Some of the fixtures you mentioned, on the other hand, are fine, they're used for testing certain behavior of the repo builder tooling, not in "real" deploys (same for e.g. the dev center generator fixtures).

I updated a ton of others (that e.g. only did local dry-run `composer installs` for verification), too, to bring them to up to heroku-24 state, and restored Composer v1 specific tests (which is EOL and no longer available, but we want to ensure it fails as expected).

I did spot two leftover tests however in `platform_spec.php` that use the `:requires_php_on_stack` RSpec filter, one is obsolete but the other needs updating.

edmorley (approved):

PR #795, opened 2025-05-07
Post-heroku-20 cleanups

dzuelke:
Forgot to remove the `libcassandra` formula which is now obsolete (because the extension does not support PHP 8).

Also a little more cleanup to some build formulae; `apcu` and `phalcon` are now "vanilla" PECL builds.

Plus other small fixes from before the `heroku-20` removal: stray `:focused` filter in a test (ignored in CI), some `expect()` arg order in tests for better failure messages, and redundant `s5cmd` install for CI.

GUS-W-14707054


edmorley (approved):
> stray :focused filter in a test (ignored in CI)

In case it's useful to you, the `rubocop-rspec`'s [RSpec/Focus](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecfocus) lint often helps me catch these:
https://github.com/heroku/heroku-buildpack-python/blob/819fb61774b66abddbd998db5c44f12d0d03afb1/.rubocop.yml#L22-L25

PR #796, opened 2025-05-07
Platform package repository snapshots (and support for build metadata in package versions)

dzuelke:
This handles packages such as "`php-8.4.6+build2`", which will allow re-building of existing packages without overwriting them in a published repository (a pre-requisite for stable repository snapshots).

Because semver ignores build metadata for version comparison, if multiple packages with the same version but different build metadata exists, `mkrepo.sh` will only include the "highest" build metadata variant (sorted using "natural", version-like sorting, so "`+build11`" is higher than "`+build9`" etc).

For snapshots, a hash is calculated based on the sorted list of formulae in `support/build/`. If a formula is renamed or added, the hash changes. During repository upload, `mkrepo.sh` will create a `packages-${hash}.json` in addition to `packages.json` (which still serves as the "bleeding edge" repository, as before).

In `bin/compile`, the same hash is used to calculate a repository URL.

Together with the changes for build metadata variants in package formulae versions described above, this allows stable repository snapshots, including re-builds of existing published package versions, that are not picked up by published buildpack releases, since they would reference a different snapshot hash in the URL to `packages-${hash}.json`.

CI now requires platform repo snapshots to be present on the develop bucket/prefix for pull request runs. This still allows manual `workflow_dispatch` runs without a snapshot, e.g. during development/debugging. Tags and main run against the default repo, as usual, which now uses snapshots. The Prepare Release workflow enforces the presence of a snapshot on the stable bucket/prefix as an additional measure.

GUS-W-18461573
GUS-W-18463967


Malax (approved):
Looks good to me, I'm missing some context for the finer details but the code has comments to explain these. But I did not verify that the comments were correct.

The only thing that stood out to me are the inconsistencies in the test fixture `.json` files when it comes to trailing newlines. They probably should have them. I'll approve this PR nonetheless to avoid unnecessary re-review loops. Please consider fixing those before merging. :)

dzuelke:
Yes, well spotted @Malax, I tidied those up (also in any other fixtures that were missing them, except for all the `expected_platform_composer.json`s, which are verbatim generated like that without a trailing newline).

dzuelke:
Sync (only to create snapshots on `dist-${STACK}-stable/`): https://github.com/heroku/heroku-buildpack-php/actions/runs/14906653805

support/build/README.md
```diff
@@ -371,6 +371,14 @@ Additional dependencies can be expressed as well; for example, if an extension r
...
 
+#### Package version
+
+The version of a package must conform to [Composer's version specification](https://getcomposer.org/doc/articles/versions.md) and [Semantic Versioning](https://semver.org).
+
+When updating the build for an existing, published version of a package (e.g. with updated compile options), it is recommended to use *build metadata* in the version string of the updated package in order to distinguish it from the existing, published version. For example, `php-8.4.6` would become `php-8.4.6+build2`.
+
+Per Semver specifications, any build metadata in a version string is ignored during version comparison. This means it could not be guaranteed that `php-8.4.6+build2` is picked, if it exists in the same repository as `php-8.4.6`. For this reason, when [(Re-)generating Repositories](re-generating-repositories), only the package/version combination with the "highest" (compared using lenient version string comparison) build metadata will be included, meaning that in the example above, `php-8.4.6` would not be included in the generated repository, only `php-8.4.6+build2`.
```
schneems, line 380: This link is broken:

```
$ curl -I -s https://github.com/heroku/heroku-buildpack-php/blob/main/support/build/re-generating-repositories | grep HTTP
HTTP/2 404
$ curl -I -s https://github.com/heroku/heroku-buildpack-php/blob/0f0a5075ae34b471ca17bbc2c05bb88e8c01ca3e/support/build/re-generating-repositories | grep HTTP
HTTP/2 404
```

I think you need a `#` so it's an anchor link. That's what the link in the `### Repository Snapshots` section uses.

PR #805, opened 2025-05-15
Fix buildpack.toml exclusion list

dzuelke:
Accidentally excluded `support/build/_util/formulae-hash.sh`, which is used in `bin/compile` to calculate the repo snapshot hash.

GUS-W-18463967


runesoerensen (approved):

buildpack.toml
```diff
@@ -6,7 +6,9 @@ name = "PHP"
 		".github/",
 		".gitignore",
 		".rspec_parallel",
-		"support/build/_*/",
+		"support/build/_conf/",
```
edmorley, line 9: @dzuelke This was merged without a changelog entry, since this repo doesn't have the check-changelog check:
https://github.com/heroku/heroku-buildpack-php/blob/main/CHANGELOG.md#v267---2025-05-15

PR #811, opened 2025-07-01
Fix spawned background processes (e.g. ScoutAPM core-agent) causing builds to hang forever

dzuelke:
If a background process spawns during the build, e.g. an APM agent such as ScoutAPM, it inherits open file descriptors. At the end of the build, the parent process (the buildpack, or whatever the process leader is for the child process) will wait for that file descriptor to be closed, even if it is closing it on the parent side. The child process, however, may never close it, since it didn't know about it upon process start, leading to the build hanging forever.

GUS-W-18850207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### FIX
+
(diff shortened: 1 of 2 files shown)


edmorley:
@dzuelke So just to confirm, this fixes a regression from #803 right?

PR #816, opened 2025-07-07
Test only, do not merge: hatchet gitref

dzuelke:


edmorley:
Converting to draft, otherwise this appears in the review queue

PR #820, opened 2025-07-24
Colorization and indentation cleanup and fixes

dzuelke:
Notices, warnings and errors are currently printed in yellow or red, but the ANSI codes used are supposed to wrap to following lines. This causes very inconsistent output across environments; for instance, a `git push` will have its `remote: ` prefix highlighted in yellow or red for subsequent lines of a warning or error, while the build logs in Heroku Dashboard will not colorize these lines at all (because the ANSI codes do not re-start on the next line):

Screenshot 2025-07-24 at 16 58 56

There are other subtle flaws, too; for instance, empty lines are currently swallowed by the build logs view in Heroku Dashboard (which we can prevent by making them non-empty).

This now allows passing an indentation level to any of the message helpers, as well as a prefix (useful for upcoming changes which need this).

For the `indent` helper, the "do not indent first line" (used in various places) case is now a lot less confusing, since it's a flag argument instead of a magic empty-or-not-empty string.

Color is now handled correctly at line boundaries, meaning warning or error colors do not leak to the next line (where a wrapper, such as Git, might be printing its own prefix).

As a result, colors are now consistently and correctly printed on the CLI and e.g. in the build logs in Heroku Dashboard.

Smaller fixes include warning summaries at the end in case of an error (now printed in yellow which is easier to "spot"), and the list of available runtime versions (indented cleanly).

### `git push` old vs new:

terminal git push old
terminal git push new

### Dashboard old vs new:

Heroku Dashboard build log old
Heroku Dashboard build log new

GUS-W-19126260

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -520,7 +520,12 @@ else
 		)
 		For reference, the following runtimes are currently available:
 		
-		PHP:  $(platform-composer show --available heroku-sys/php 2>&1 | sed -n 's/^versions : //p' | fold -s -w 58 || true)
+		PHP: $(
+			platform-composer show --available heroku-sys/php 2>&1 |
+			sed -n 's/^versions : //p' | # composer lists them like "versions : 1.2.3, 1.2.4"
+			fold -b -s -w 59 | # 59 characters at a time, wrap on whitespace
+			indent -i5 --no-first-line-indent || true # indent all lines except first to line up with "PHP: "
+		)
 		
 		Please verify that all requirements for runtime versions in
 		'$COMPOSER_LOCK' are compatible with the list above, and ensure
--- a/bin/compile
+++ b/bin/compile
@@ -496,25 +496,26 @@ else
 		
 		$install_log
 		$(
-			$composer_lock_outdated && cat <<-EOF2
-				$(echo -e "\033[1;33m")
+			# via sed, make every line yellow at the start and re-set to red at the end
+			$composer_lock_outdated && cat <<-EOF2 | sed -e $'s/^/\033[1;33m/' -e $'s/$/\033[1;31m/'
+				
 				A possible cause for this error is your '$COMPOSER_LOCK' file,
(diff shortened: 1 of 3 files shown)


bin/util/common.sh
```diff
@@ -2,106 +2,255 @@
...
 _captured_warnings_file=$(mktemp -t heroku-buildpack-php-captured-warnings-XXXX)
 
+# Get config from caller's args ("$@") and set results into return variables.
+# The following flags and options can be given:
+# --no-first-line-indent
+# --rjust (to right-justify the computed prefix, i.e. pad with spaces on left)
+# -p PREFIXSTRING (e.g. " !", defaults to "")
+# -i INDENTWIDTH (e.g. 7, defaults to 7)
+# It populates the following variables, if non-empty at call time:
+# - OPTIND (for re-settting "$@" using 'shift')
```
colincasey, line 12:
```suggestion
# - OPTIND (for re-setting "$@" using 'shift')
```

bin/compile
```diff
@@ -496,30 +496,36 @@ else
...
+				
 			EOF2
 		)
 		For reference, the following runtimes are currently available:
 		
-		PHP:  $(platform-composer show --available heroku-sys/php 2>&1 | sed -n 's/^versions : //p' | fold -s -w 58 || true)
+		PHP: $(
+			platform-composer show --available heroku-sys/php 2>&1 |
+			sed -n 's/^versions : //p' | # composer lists them like "versions : 1.2.3, 1.2.4"
+			fold -b -s -w 59 | # 59 characters at a time, wrap on whitespace
```
colincasey, line 526: Why did this change from `-w 58` to `-w 59`?

dzuelke: Because there were two spaces after "PHP:" before (legacy from back when we had HHVM as well, for which the versions were listed on the following lines), so we now have one more character to work with :)

PR #825, opened 2025-07-31
Late July 2025 PHP updates

dzuelke:
Runtimes, extensions and packages.

GUS-W-17779557
GUS-W-17779558
GUS-W-17779559


dzuelke:
Sync: https://github.com/heroku/heroku-buildpack-php/actions/runs/16658087821

PR #846, opened 2025-09-30
Late September 2025 PHP updates

dzuelke:
GUS-W-17779563
GUS-W-17779564
GUS-W-17779565


dzuelke:
Sync: https://github.com/heroku/heroku-buildpack-php/actions/runs/18145795809

PR #917, opened 2026-03-27
Fix PHP extension loading when custom php.ini is provided

fagianijunior:
**Description:**
This PR addresses an issue where PHP extensions might not load correctly when a custom php.ini file is provided via the -c option in the PHP-FPM command. The problem occurs because a custom php.ini may not define the extension_dir directive, leading to extensions not being found even if they are installed in the correct location.

**Changes Made:**
Added logic to retrieve the correct extension directory using php-config --extension-dir.
Modified the fpm_command array in both *heroku-php-apache2* and *heroku-php-nginx* to include the -d extension_dir=$php_extension_dir parameter.
This ensures that the extension directory is always set correctly, regardless of whether a custom php.ini is used or not.

**Files Modified:**
bin/heroku-php-apache2
bin/heroku-php-nginx

**Testing:**
Verified that PHP-FPM starts correctly with custom php.ini files.
Confirmed that extensions load properly in both Apache2 and Nginx configurations.

This fix improves compatibility and reliability for users providing custom PHP configurations in Heroku deployments.

### DIFF
--- a/bin/heroku-php-apache2
+++ b/bin/heroku-php-apache2
@@ -389,7 +389,10 @@ echo -e "\n[global]\nlog_level = notice" >> "$fpm_config_tmp"
 
 # build command string arrays for PHP-FPM and HTTPD
 # we're using an array, because we need to correctly preserve quoting, spaces, etc
-fpm_command=( php-fpm --prefix "$php_server_root" --pid "$fpm_pidfile" --nodaemonize -y "$fpm_config_tmp" ${php_config:+-c "$php_config"} )
+# when a custom php.ini is given via -c, it replaces the buildpack's php.ini, which may not define extension_dir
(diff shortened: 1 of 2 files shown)


edmorley (commented):
Thank you for the PR, and sorry for the delay in replying. 

Given the wide variety of app configurations used in the wild, even seemingly innocuous changes can end up breaking apps that relied upon a specific behaviour, so in general we have to be very confident about a change before being able to accept it.

And unfortunately our previous PHP buildpack maintainer no longer works at Heroku, and the remaining language owners aren't as familiar with PHP / this buildpack which makes it harder for us to review contributed changes like these.

As such for now the best way for you to use this is likely via a fork of the buildpack.

PR #922, opened 2026-04-14
April 2026 PHP platform updates

schneems:
GUS-W-20573228
GUS-W-20573227
GUS-W-20573226


support/build/packages/ext-blackfire-2026.4.0_php-8.1
edmorley, line 1: (observation only) I noticed the versioning scheme has changed - so I confirmed the probe/extension appears to have now switched to the newer calver scheme used by the CLI by checking the versions here:
https://blackfire.io/api/v1/releases

...and looks good to me :-)

(Documented at https://github.com/heroku/languages-team/blob/main/languages/php/platform.md#what-to-release-and-when)

edmorley (approved):
Thank you!

support/build/packages/composer-2.9.7
edmorley, line 1: Btw I noticed that Composer 2.2.x also had an update (it seems Composer 2.2 it the LTS release, so they still backport fixes to that):
https://github.com/composer/composer/releases/tag/2.2.27

(None of the other Composer 2.x minors had a new patch update - only 2.9.x which is in this PR already, and 2.2.x as mentioned above)

schneems: Updated, please 👀

edmorley: Looks good - thank you!

edmorley (approved):

PR #929, opened 2026-04-15
Fix bootstrap_url_filepath path resolution when invoked via relative path

### LINKED ISSUE #928: `make run` development workflow broken after #907
It fails with:

```
$ make run
Running buildpack using: STACK=heroku-24 FIXTURE=test/fixtures/default

~ Detect: PHP

~ Compile:
 !     
 !     WARNING: No 'composer.json' found!
 !     
 !     Your project only contains an 'index.php', no 'composer.json'.
 !     
 !     Using 'index.php' to declare app type as PHP is deprecated and
 !     may lead to unexpected behavior.
 !     
 !     Please consider updating your codebase to utilize Composer and
 !     modern dependency management in order to benefit from the latest
 !     PHP runtimes and improved application performance, as well as
 !     control over the PHP versions and extensions available.
 !     
 !     For an introduction to dependency management with Composer and
 !     how to get the most out of PHP on Heroku, refer to the docs at
 !     https://getcomposer.org/doc/00-intro.md and
 !     https://devcenter.heroku.com/articles/getting-started-with-php
 !     
-----> Bootstrapping...
find: './bin/../support/build/packages': No such file or directory
 !     WARNING: Could not find a PHP version for bootstrapping!
 !     
 !     ERROR: Failed to download minimal PHP for bootstrapping!
 !     
 !     This is most likely a temporary internal error. If the problem
 !     persists, make sure that you are not running a custom or forked
 !     version of the Heroku PHP buildpack which may need updating.
 !     
 !     REMINDER: the following warnings were emitted during the build;
 !     check the details above, as they may be related to this error:
 !     - No 'composer.json' found!
 !     - Could not find a PHP version for bootstrapping!
 !     

~ Report:
{
  "bootstrap.duration": 7.684482,
  "duration": 7.689626,
  "failure_detail": "php-min",
  "failure_reason": "bootstrap.download",
  "open_timers": "__main__,bootstrap"
}
make: *** [run] Error 1
```

This appears to be due to the new `formulae-hash.sh` script here not handling `bin/compile` having been called by a relative path:
https://github.com/heroku/heroku-buildpack-php/blob/5efec4875e5c88a6e5c495f056cf84ab98e70245/bin/compile#L262

The `make run` target has always used a relative path:
https://github.com/heroku/heroku-buildpack-php/blob/8ec06d34745111b3eb17d8da71c8bbaf9aeb0c07/Makefile#L29

The `formulae-hash.sh` script was added in https://github.com/heroku/heroku-buildpack-php/pull/796.

And also the fact that `make run` isn't tested in CI. See Python's CI for an example of testing this (which would have prevented this DX regression):
https://github.com/heroku/heroku-buildpack-python/blob/b8216105773cb602c576eef22d6afe8200b507e1/.github/workflows/ci.yml#L66-L76

schneems:
Use $bp_dir (already resolved to an absolute path on line 21) instead of
$(dirname "$BASH_SOURCE") which produces a relative path when bin/compile
is called as ./bin/compile. After `cd $build_dir`, the relative path no
longer resolves to the buildpack's support/build/packages directory.

Fixes https://github.com/heroku/heroku-buildpack-php/issues/928

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,13 @@ jobs:
           ruby-version: "${{ env.setup_ruby_ruby_version }}"
       - name: Execute tests
         run: bundle exec rspec test/spec/cgroup_spec.rb
+  container-test:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Run buildpack using default app fixture
+        run: make run
   integration-test:
     runs-on: ubuntu-24.04
     strategy:
--- a/bin/compile
+++ b/bin/compile
@@ -285,7 +285,7 @@ if [[ "${HEROKU_PHP_PLATFORM_REPOSITORIES}" == *" - "* ]]; then
 fi
 
 bootstrap_url_filepath() {
-	local bootstrap_scan_dir=$(dirname "$BASH_SOURCE")/../support/build/packages
+	local bootstrap_scan_dir="$bp_dir/support/build/packages"
 	
 	# 1. Find
 	#   - in $bootstrap_scan_dir


edmorley (approved):
Thank you!

PR #946, opened 2026-05-13
Allow PHP 8.5 for default PHP version selection

runesoerensen:
## Summary

- Bumps the upper bound of `HEROKU_PHP_DEFAULT_RUNTIME_VERSION` in `bin/compile` from `<8.5` to `<8.6`. Apps without an explicit `php` requirement (in `composer.json` or any userland dependency) are now eligible to resolve to PHP 8.5.
- Updates the `defaultphp` platform-generator test fixture (`ENV`, `expected_platform_composer.json`, `expected_stderr`) to match.
- Updates `expected_default_php` in `test/spec/spec_helper.rb` from `8.4` to `8.5` so the `php_default_spec` integration test asserts the correct series. (This mirrors the precedent set in `7ece0b14` "PHP/8.4.1", which bumped both the default range and `expected_default_php` together.)

GUS-W-20141331


edmorley:
> ## Test plan
> 
>     * [ ]  CI passes (platform generator + integration tests)

Do we feel these kind of entries in the PR description add value? I presume they are LLM generated? Saying "we'll test this PR by making sure CI passes" just feels a bit redundant - that's the whole point of branch protection rules required status checks etc.

edmorley (approved):
Am I correct in remembering that the PHP CNB doesn't need this change since it already defaults to the latest PHP version? If so, do we know why the behaviour differs here and/or should we unify them at some point?

runesoerensen:
> > ## Test plan
> > ```
> > * [ ]  CI passes (platform generator + integration tests)
> > ```
> 
> Do we feel these kind of entries in the PR description add value? I presume they are LLM generated? Saying "we'll test this PR by making sure CI passes" just feels a bit redundant - that's the whole point of branch protection rules required status checks etc.

No we don't, and I'm not sure how many times I have to tell Claude not to include it in PR descriptions :) I've removed it, and will try and find a better way to drop these consistently.

runesoerensen:
> Am I correct in remembering that the PHP CNB doesn't need this change since it already defaults to the latest PHP version?

Yes as far as I can tell the CNB doesn't need this change.

> If so, do we know why the behaviour differs here and/or should we unify them at some point?

In [`buildpacks/php/src/package_manager/composer.rs#L307-L319`](https://github.com/heroku/buildpacks-php/blob/main/buildpacks/php/src/package_manager/composer.rs#L307-L319), when no `php` requirement exists anywhere (root or transitively), the CNB inserts `heroku-sys/php: "*"` and lets Composer pick the latest available version from the platform repo. There's no upper bound, so a new minor like 8.5 becomes the default automatically as soon as it's published to the repo.

Definitely seems worth unifying approaches... Not sure what the best solution is though, but leaning toward dropping the upper bound like the CNB now that we have a more robust and stable repo/bucket solution? Seems dev/test failures were the motivation for introducing it here initially https://github.com/heroku/heroku-buildpack-php/pull/670

edmorley:
> Seems dev/test failures were the motivation for introducing it here initially #670

Ah yes, seems like the motivation for it no longer exists after the switch to the S3 bucket manifest snapshotting approach :-)

runesoerensen:
> > Seems dev/test failures were the motivation for introducing it here initially #670
> 
> Ah yes, seems like the motivation for it no longer exists after the switch to the S3 bucket manifest snapshotting approach :-)

I've filed a ticket to follow-up on this :)

PR #971, opened 2026-07-12
Add ShellCheck linting to CI

### LINKED ISSUE #934: Lint Bash scripts in CI using shellcheck
This buildpack doesn't currently have any linting of Bash scripts using shellcheck, unlike some of our other buildpacks. And in particular, this buildpack makes heavy use of advanced Bash features (as seen by the fact that it's the only Bash buildpack so far running into Bash 5.3 compatibility issues), so it seems linting is even more important here.

We should add this to CI - initially with any failing errors/warnings disabled so that we can at least prevent further lint regressions while we clean up the existing usages.

See Python for an example of doing this:

https://github.com/heroku/heroku-buildpack-python/blob/b8216105773cb602c576eef22d6afe8200b507e1/.github/workflows/ci.yml#L18-L35

https://github.com/heroku/heroku-buildpack-python/blob/b8216105773cb602c576eef22d6afe8200b507e1/.shellcheckrc#L1-L2

https://github.com/heroku/heroku-buildpack-python/blob/b8216105773cb602c576eef22d6afe8200b507e1/Makefile#L12-L15

oiahoon:
## Summary
- add a ShellCheck configuration that enables optional checks while baselining current findings
- add a Make target that checks shell entry points and helpers while excluding PHP executables
- run the lint target in a standalone CI job

The disabled codes are the exact set reported by the current main branch, keeping this change focused on enabling enforcement rather than modifying existing buildpack behavior.

## Testing
- make lint-scripts
- bundle exec rspec test/spec/cgroup_spec.rb (288 examples, 0 failures)
- workflow YAML structure validation
- git diff --check

Fixes #934


Makefile
```diff
@@ -16,6 +16,9 @@ COMPILE_FAILURE_EXIT_CODE ?= 1
 # Converts a stack name of `heroku-NN` to its build Docker image tag of `heroku/heroku:NN-build`.
 STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
 
+lint-scripts:
+	@git ls-files -z --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' ':!:*.php' | xargs -0 shellcheck --check-sourced --color=always
```
edmorley: Claude says:
---

**Coverage gap: `support/build/formulae/*` and `support/build/packages/*` are silently not linted.** — `Makefile` `lint-scripts` target

The glob `'bin/*' '*/bin/*' '*.sh'` misses ~168 bash-shebang scripts in this repo that have neither a `.sh` extension nor a `bin/` path — notably the entire binary-build toolchain (`support/build/formulae/php`, `.../formulae/apache`, `support/build/packages/php-8.4.23`, etc.). These are hand-maintained bash scripts, and linting them surfaces codes **not** in the baseline (`SC2038`, `SC2115`, `SC2156`, `SC2231` — several of which are actual-bug-class checks like unquoted `find | xargs` and possibly-wrong `test`).

.github/workflows/ci.yml
```diff
@@ -30,6 +30,13 @@ env:
   setup_ruby_ruby_version: "4.0"
 
 jobs:
+  lint:
```
edmorley, line 33: The new lint job is failing in CI:
https://github.com/heroku/heroku-buildpack-php/actions/runs/29212762748/job/86815906053?pr=971

edmorley (commented):
Hi! Thank you for the PR - I've left some comments :-)

Makefile
oiahoon: gotcha

.github/workflows/ci.yml
oiahoon: Addressed in 891132bf. SC2317 is now included in the explicit baseline, and make lint-scripts passes locally across all selected paths.

Makefile
oiahoon: Addressed in 891132bf. The lint target now explicitly includes support/build formulae and package scripts, with their existing findings recorded in the baseline.

edmorley:
Triggered CI from this repo via #975 so that the integration tests can run (secrets aren't available to CI run from forks).

edmorley (approved):
Thank you for the PR!

Over time we can now start to fix and re-enable some of the disabled Shellcheck rules (for the rules that make sense and aren't too obtrusive given legacy codebase).

PR #981, opened 2026-08-18
Replace /dev/stderr with standard stream channel

nerg4l:
https://github.com/gliderlabs/herokuish uses an unprivileged user to run the buildpack. Unfortunately, this causes writing to `/dev/stderr` to fail with permission denied.

Back in march I tried to resolve this in https://github.com/gliderlabs/herokuish/issues/1818 but my attempt uses TTY which is not available during `git push`.

After a lot of try and error, it seems the simplest would be to update the code here with an equivalent solution that uses the standard stream channel instead of `/dev/stderr`. Based on my tests, this implementation behaves the same and now supports herokuish.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -454,7 +454,7 @@ status "Preparing platform package installation..."
 HEROKU_PHP_DEFAULT_RUNTIME_VERSION="^8.1.0 <8.6"
 export HEROKU_PHP_DEFAULT_RUNTIME_VERSION
 # extract requirements from composer.lock
-{ platform_stderr=$(/app/.heroku/php-min/bin/php $bp_dir/bin/util/platform.php ${have_custom_platform_repos:+"--list-repositories"} "$bp_dir/support/installer/" $HEROKU_PHP_PLATFORM_REPOSITORIES 2>&1 >$build_dir/.heroku/php/composer.json | tee >(indent > /dev/stderr)); } || {
+{ platform_stderr=$(/app/.heroku/php-min/bin/php $bp_dir/bin/util/platform.php ${have_custom_platform_repos:+"--list-repositories"} "$bp_dir/support/installer/" $HEROKU_PHP_PLATFORM_REPOSITORIES 2>&1 >$build_dir/.heroku/php/composer.json | tee >(indent >&2)); } || {
 	code=$?
 	if (( code == 3 )); then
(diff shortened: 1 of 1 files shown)


schneems:
This was introduced in https://github.com/heroku/heroku-buildpack-php/pull/852. I can't find a reason given at the time. I think this is functionally equivalent for our needs. My best guess is that reading`/dev/stderr` is more self-documenting than `>&2` but honestly, it's just a meh guess since we use `>&2` all over the place in all the bash buildpacks. I don't have any strong objections. We'll need to pull this in to CI and write a changelog for it.

edmorley:
I'm happy to try and see if we can merge this (post-moratorium).

> We'll need to pull this in to CI and write a changelog for it.

PRs by default have the "Maintainers are allowed to edit this pull request." option enabled, which means we should be able to:
1. Push extra commits to the PR
2. Mirror that PR's branch to this repo
3. Open a draft PR from that branch (which runs CI and updates the status checks for both PRs, since they share the same SHA)
4. Approve/merge this PR and close out the draft PR.

And if we add support for a manual `workflow_dispatch` event trigger for CI workflows, then we can even skip step (3) in the future. (I'll open PRs to do that now.)

nerg4l:
@edmorley That would be great. While checking the code, I also found a reference to `/dev/stdin`, but I did not investigate it because at them moment, I don't face any issues with that reference. https://github.com/heroku/heroku-buildpack-php/blob/6a7e4200ff5ba56e95c7b4cbc42bb0907e107b53/bin/util/failures.sh#L151

I checked the "Maintainers are allowed to edit this pull request." when I created the PR, and I just confirmed that it is indeed checked.

edmorley:
> And if we add support for a manual `workflow_dispatch` event trigger for CI workflows, then we can even skip step (3) in the future. (I'll open PRs to do that now.)

Turns out PHP already has this:
- https://github.com/heroku/heroku-buildpack-php/blob/6a7e4200ff5ba56e95c7b4cbc42bb0907e107b53/.github/workflows/ci.yml#L8
- https://github.com/heroku/buildpacks-php/blob/1d398ea2e9853d45a00679b28a9b5da5752664d2/.github/workflows/ci.yml#L9

...but I'll add it for the other buildpack repos now.

edmorley (approved):

